PR #20, opened 2022-01-07
Add the `rr_invoke.jl` script as a convenient wrapper for running the Yggdrasil `rr` binary

DilumAluthge:


DilumAluthge:
@Keno I thought that instead of having users run e.g. `/home/chriselrod/.julia/artifacts/dbbbb27d561acd27fe742e87037103ad4fa87af3/bin/rr replay .`, it might be a little nicer to give them this small wrapper script and have them run `julia rr_invoke.jl replace .`.

PR #32, opened 2022-01-28
Add hooks directory, test a hook

brenhinkeller:
The `.buildkite/hooks` directory should allow us to implement the custom `checkout` hook we need


brenhinkeller:
Now to see what it takes to get buildkite to run on this PR...

PR #33, opened 2022-01-28
Add hooks directory, test a hook

brenhinkeller:
The .buildkite/hooks directory should allow us to implement the custom checkout hook we need

### DIFF
--- a/.buildkite/hooks/post-checkout
+++ b/.buildkite/hooks/post-checkout
@@ -8,6 +8,6 @@ ls -alh
 echo 'remove current contents of build directory'
 rm -rf ./*
 echo 'Clone julia, main branch only:'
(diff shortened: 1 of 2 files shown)


DilumAluthge:
@staticfloat It would be nice (but not a dealbreaker) if we could use the "repo mirrors experiment" to cache these clones of JuliaLang/julia, just to speed up that part.

staticfloat:
Agreed; I told Brenhin to forge ahead with pure `git clone` for now, and we'll figure out how to hook into the git mirrors afterwards.

staticfloat:
Merging this in as a good first POC.  Will follow up with a PR that strips out many of the pipelines so we can iterate faster for a bit.

PR #41, opened 2022-02-27
Disable most pipelines so we can quickly iterate for now

staticfloat:


brenhinkeller (approved):
LGTM!

brenhinkeller:
(the messages and cleanup, that is, not sure if I understand the pipelines really 😆 )

PR #44, opened 2022-03-09
Enable macos packaging and tests

gbaraldi:
Depends on #42. I have no idea if this is working or not. This is a rebase on [this PR](https://github.com/JuliaLang/julia/pull/43529) but with a similar structure to what is on #42


staticfloat:
I just merged a big re-organization, sorry about that.  You'll probably need to rebase this significantly.

gbaraldi:
I was already building it on top of the branch of the merged PR, so it shouldn't be too hard.

gbaraldi:
I was already building it on top of the branch of the merged PR, so it was a simple merge.

staticfloat:
Can you enable aarch64 as well?

pipelines/main/platforms/build_macos.arches
```diff
@@ -0,0 +1,13 @@
+# OS       TRIPLET                 ALLOW_FAIL    ARCH_ROOTFS    MAKE_FLAGS     TIMEOUT       ROOTFS_TAG    ROOTFS_HASH
```
staticfloat, line 1: No need for `ARCH_ROOTFS`, `ROOTFS_TAG` or `ROOTFS_HASH` here.  Those columns can be deleted.

staticfloat: `ARCH_ROOTFS` should only be used with sandbox; any other uses of the arch should parse it out from the triplet, [which is done by `calc_version_envs.sh`](https://github.com/JuliaCI/julia-buildkite/blob/main/utilities/calc_version_envs.sh#L4)

pipelines/main/platforms/build_macos.yml
```diff
@@ -0,0 +1,51 @@
...
+      - label: ":macos: build ${TRIPLET?}"
+        key: "build_${TRIPLET?}"
+        plugins:
+          - JuliaCI/external-buildkite#v1:
+              version: ".buildkite-external-version"
+              repo_url: "https://github.com/JuliaCI/julia-buildkite"
+          - JuliaCI/julia#v1:
+              # Drop default "registries" directory, so it is not persisted from execution to execution
+              persist_depot_dirs: packages,artifacts,compiled
+              version: '1.6'
```
staticfloat, line 13: This is not needed; you can delete it.

pipelines/main/platforms/test_macos.yml
```diff
@@ -0,0 +1,65 @@
...
+        depends_on:
+          - "build_${TRIPLET?}"
+        plugins:
+          - JuliaCI/external-buildkite#v1:
+              version: ".buildkite-external-version"
+              repo_url: "https://github.com/JuliaCI/julia-buildkite"
+          - JuliaCI/julia#v1:
+              # Drop default "registries" directory, so it is not persisted from execution to execution
+              persist_depot_dirs: packages,artifacts,compiled
+              version: '1.6'
```
staticfloat, line 15: This is not needed, you can delete it.

pipelines/main/platforms/build_macos.yml
```diff
@@ -32,7 +32,7 @@ steps:
           ./julia -e 'using InteractiveUtils; InteractiveUtils.versioninfo()'
 
           echo "--- Create build artifacts"
-          make --output-sync -j 8 binary-dist ${MAKE_FLAGS?}
+          gmake --output-sync -j 8 binary-dist ${MAKE_FLAGS?}
```
staticfloat, line 31: I don't think you want `gmake`.  This isn't FreeBSD, Mac's have GNU Make by default:

```
$ make --version
GNU Make 3.81
```

gbaraldi: I tried that at first https://buildkite.com/julialang/julia-buildkite/builds/112#9229a63f-d883-4f03-8ccc-08ef8e751148 but `--output-sync` is only available on newer make. Make 3.81 is from 2006 😄 .

gbaraldi: I guess I can just make without it, it's not that big of an issue. 
It seems like aarch64 has it and x86 doesn't for some reason.

gbaraldi:
Yeah, I was very very confused, I was like wow, aarch64 is building very nice, then it was on the x86 builder 😄 .

staticfloat:
ARCH_ROOTFS is supposed to refer to the architecture of the rootfs image that we're downloading, which isn't happening on these machines, so I'd rather name it just `ARCH`.

gbaraldi:
Isn't there an ARCH variable already? That `calc_version_envs.sh` sets?

staticfloat:
Yes, but that variable is not available at YAML parse time, sadly.

gbaraldi:
Fair enough, I guess adding to the .arches file won't hurt. Not sure why the aarch64 build didn't work. I donwloaded it and it gave me broken dylib errors 🤔 . Will try to see if current master actually works on the m1.

staticfloat:
The aarch64 tester is failing because the binaries are not codesigned.  I've updated our testing fork of Julia to include the latest fixes from Mosè for that, so the next build you push should have more luck there.

This is looking pretty good!  I think we're ready to embed the codesigning toolchain, so I went ahead and did that for you.  Let's have the workflow be:

* build
  - generates `.tar.gz`
* test
  - waits for build
  - downloads `.tar.gz` and extracts it
  - re-signs the binaries with ad-hoc signatures (because the build doesn't have a codesigning key, so I don't think julia will run without that)
  - runs tests
* upload
  - waits for test to complete
  - downloads `.tar.gz` and extracts it
  - re-signs the binaries with an actual codesigning key
  - generates, signs and notarizes `.dmg`
  - uploads tarball, tarball GPG key, and signed `.dmg`

staticfloat:
The aarch64 tester is failing because the binaries are not codesigned.  I've updated our testing fork of Julia to include the latest fixes from Mosè for that, so the next build you push should have more luck there.

This is looking pretty good!  I think we're ready to embed the codesigning toolchain, so I went ahead and did that for you.  Let's have the workflow be:

* build
  - generates `.tar.gz`
* test
  - waits for build
  - downloads `.tar.gz` and extracts it
  - re-signs the binaries with ad-hoc signatures (because the build doesn't have a codesigning key, so I don't think julia will run without that)
  - runs tests
* upload
  - waits for test to complete
  - downloads `.tar.gz` and extracts it
  - re-signs the binaries with an actual codesigning key
  - generates, signs and notarizes `.dmg`
  - uploads tarball, tarball GPG key, and signed `.dmg`

gbaraldi:
That sounds great to me, thanks :). I will see if later tonight or maybe tomorrow I can get to that. 
It's quite incredible how much faster the m1 builds, it's almost 3x faster than the normal x86.
Getting the CI for CI working here was awesome for developing this., thanks elliot and @brenhinkeller .

staticfloat:
> It's quite incredible how much faster the m1 builds, it's almost 3x faster than the normal x86.

The x86_64 machine is pretty old: `Intel(R) Core(TM) i5-4308U CPU @ 2.80GHz`, it's a personal machine of mine that I'm using for testing.  The "true" buildbots that we tend to use are newer: `Intel(R) Core(TM) i7-8700B CPU @ 3.20GHz`, so we should get significantly better performance out of those.  Those new ones also have 12 cores instead of 4, so I think we can parallelize builds and test suites such that things run a little faster on them!  :D

pipelines/main/platforms/test_macos.yml
```diff
@@ -23,6 +23,11 @@ steps:
           echo "--- Extract build artifacts"
           tar xzf "$${UPLOAD_FILENAME}" "$${JULIA_INSTALL_DIR}/"
 
+          echo "--- Ad-hoc sign for testing"
+          MACHO_FILES=$(find "$${JULIA_INSTALL_DIR}" -type f -perm -0111 | cut -d: -f1)
+          for f in ${MACHO_FILES}; do
+            codesign -s - --option=runtime -vvv --timestamp --deep --force "${f}"
```
staticfloat, line 29: If possible, let's make use of https://github.com/JuliaLang/julia/blob/master/contrib/codesign.sh here

gbaraldi: Do we have access to it directly from the julia repo or do I add it to utilities?

gbaraldi:
Why does adhoc signing not work here 🤔 why does apple make this so more difficult than it should be.

DilumAluthge:
@staticfloat @gbaraldi Now that Buildkite has a built-in matrix feature, what do you guys think of trying it out for this PR?

Docs: https://buildkite.com/docs/pipelines/build-matrix

staticfloat:
We may be able to do that, as long as we don't have to use the matrix attributes in a plugin.  (E.g. we only interpolate the values within the `command` step).

Gabriel; are you blocked on anything, or just busy?  I want to get this PR moving again, and just want to understand where things stand.

gbaraldi:
Yeah my week was a bit busier than i expected, however today I will work on it some.

pipelines/main/platforms/upload_macos.yml
```diff
@@ -0,0 +1,125 @@
...
+              version: "./.buildkite-external-version"
+              repo_url: "https://github.com/JuliaCI/julia-buildkite"
+      - staticfloat/cryptic#v2:
+          variables:
+            - AWS_ACCESS_KEY_ID="U2FsdGVkX184v87+NPs3j9r/JoIuOrYt4/Z4wnRdklnY17NP8C8AMZvWYLJfT9t1"
+            - AWS_SECRET_ACCESS_KEY="U2FsdGVkX1+qptnxR/Mo5jZdH8OQfflRPiQBEhjgZIiTpn8KNCJYh/Cb8xxaUWazlcM9ceOlo0InDubL+J8zdg=="
+          files:
+            - .buildkite/secrets/tarball_signing.gpg
+            - .buildkite/secrets/macos_codesigning.keychain.encrypted
+    timeout_in_minutes: ${TIMEOUT?}
```
DilumAluthge, line 26:
```suggestion
    timeout_in_minutes: 90
```

Let's see if this fixes CI, and then we can figure out the interpolation later.

gbaraldi:
That error is very informative sigh

DilumAluthge:
It seems to be coming from here: https://github.com/buildkite/interpolate/blob/07f35b4ae2515fcb0e08e9f116bdb0efd09e44d1/parser.go#L221

pipelines/main/platforms/upload_macos.yml
```diff
@@ -0,0 +1,125 @@
...
+    timeout_in_minutes: 90
+    commands: |
+      # First, get things like `LONG_COMMIT` and `SHORT_COMMIT`, etc...
+      TRIPLET="${TRIPLET?}" source .buildkite/utilities/calc_version_envs.sh
+
+      echo "--- Variables for the dmg"
+      #MACOS_CODESIGN_IDENTITY="2053E9292809B66582CA9F042B470C0929340362"
+      MACOS_CODESIGN_IDENTITY="-"
+      APP_NAME="Julia-$${MAJMIN}.app"
+      DMG_NAME="$(basename $JULIA_BINARYDIST_FILENAME8 .tar.gz).dmg"
```
DilumAluthge, line 35:
```suggestion
      DMG_NAME="$$(basename $$JULIA_BINARYDIST_FILENAME8 .tar.gz).dmg"
```

DilumAluthge:
@gbaraldi Anything that would have been `$` in the original Bash script will need to instead be `$$` in the `.yaml` file.

pipelines/main/platforms/upload_macos.yml
```diff
@@ -0,0 +1,125 @@
...
+              version: "./.buildkite-external-version"
+              repo_url: "https://github.com/JuliaCI/julia-buildkite"
+      - staticfloat/cryptic#v2:
+          variables:
+            - AWS_ACCESS_KEY_ID="U2FsdGVkX184v87+NPs3j9r/JoIuOrYt4/Z4wnRdklnY17NP8C8AMZvWYLJfT9t1"
+            - AWS_SECRET_ACCESS_KEY="U2FsdGVkX1+qptnxR/Mo5jZdH8OQfflRPiQBEhjgZIiTpn8KNCJYh/Cb8xxaUWazlcM9ceOlo0InDubL+J8zdg=="
+          files:
+            - .buildkite/secrets/tarball_signing.gpg
+            - .buildkite/secrets/macos_codesigning.keychain.encrypted
+    timeout_in_minutes: 90
```
DilumAluthge, line 26:
```suggestion
    timeout_in_minutes: ${TIMEOUT?}
```

pipelines/main/platforms/upload_macos.yml
gbaraldi: That 8 is also a typo

pipelines/main/platforms/upload_macos.yml
```diff
@@ -32,7 +32,7 @@ steps:
       #MACOS_CODESIGN_IDENTITY="2053E9292809B66582CA9F042B470C0929340362"
       MACOS_CODESIGN_IDENTITY="-"
       APP_NAME="Julia-$${MAJMIN}.app"
-      DMG_NAME="$$(basename $$JULIA_BINARYDIST_FILENAME8 .tar.gz).dmg"
+      DMG_NAME="$(basename $$JULIA_BINARYDIST_FILENAME .tar.gz).dmg"
```
DilumAluthge, line 35: I suspect that `$(...)` will trip Buildkite up and that we should instead do `$(...)`.

```suggestion
      DMG_NAME="$$(basename $$JULIA_BINARYDIST_FILENAME .tar.gz).dmg"
```

gbaraldi: I thought the double $ was for external variables, I could be wrong however

gbaraldi:
Thanks @DilumAluthge I was struggling with this for a while as you can see 😄 .

DilumAluthge:
Alright, I went through and changed almost every `$` into `$$`, and now the Buildkite parsing error is gone.

pipelines/main/platforms/upload_macos.yml
```diff
@@ -0,0 +1,125 @@
+steps:
+  - label: ":macos: upload ${TRIPLET?}"
+    key:   "upload_${TRIPLET?}"
+    # We only upload to S3 if the branch is `master` or `release-*`.
+    if: ((build.branch == "master") || (build.branch =~ /^release-/))
```
DilumAluthge, line 5:
```suggestion
    # We only upload to S3 if one of the following criteria are true:
    # 1. The branch is `master`.
    # 2. The branch is `release-*`.
    # 3. The build is a tag build AND the tag is `v*`.
    # 4. The pipeline is `julia-buildkite`.
    if: (build.branch == "master") || (build.branch =~ /^release-/) || (build.tag =~ /^v/) || (pipeline.slug == "julia-buildkite")
```

pipelines/main/platforms/test_macos.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo "--- Ad-hoc sign for testing"
+          contrib/codesign.sh "-" "$${JULIA_INSTALL_DIR}"
+
+          echo "--- Print Julia version info"
+          $${JULIA_BINARY} -e 'using InteractiveUtils; InteractiveUtils.versioninfo()'
+          echo "JULIA_CPU_THREADS is: $${JULIA_CPU_THREADS}"
+          $${JULIA_BINARY} -e '@info "" Sys.CPU_THREADS'
+
+          echo "--- Set some environment variables"
+          export OPENBLAS_NUM_THREADS=4
```
staticfloat, line 35: Hmmm, I just saw this is in Linux as well.  Why?  @DilumAluthge

pipelines/main/platforms/test_macos.yml
```diff
@@ -0,0 +1,64 @@
...
+          unset JULIA_DEPOT_PATH
+          unset JULIA_PKG_SERVER
+
+          # Make sure that temp files and temp directories are created in a location that is
+          # backed by real storage.
+          export TMPDIR="$(pwd)/tmp"
+          mkdir -p $${TMPDIR}
```
staticfloat, line 42: I think this most likely is not necessary on macOS, but it also doesn't hurt, I don't think.

pipelines/main/platforms/test_macos.yml
```diff
@@ -0,0 +1,64 @@
...
+
+          # By default, run all tests.
+          export TESTS="all LibGit2/online --ci"
+
+          export JULIA_CMD_FOR_TESTS="$${JULIA_BINARY}"
+          export NCORES_FOR_TESTS="Sys.CPU_THREADS"
+          export JULIA_NUM_THREADS=4
```
staticfloat, line 49: Let's let the agent machines choose this.

staticfloat (approved):
Overall this looks good to me.   I think we'll clean it up as necessary, but as long as it works, I say merge when green!

pipelines/main/platforms/test_macos.yml
DilumAluthge: Which line? The `OPENBLAS_NUM_THREADS`?

DilumAluthge:
Closing/reopening doesn't retrigger Buildkite.

gbaraldi:
Closed by mistake. I still need to add the keychain password so we correctly sign. 
Another thing, where in S3 should I upload the DMG.

DilumAluthge:
> Another thing, where in S3 should I upload the DMG.

@staticfloat Do they just go in the same folder as the `.tar.gz` files?

pipelines/main/platforms/test_macos.arches
```diff
@@ -0,0 +1,13 @@
+# OS       TRIPLET                 ALLOW_FAIL    ARCH          MAKE_FLAGS     TIMEOUT
+macos      x86_64-apple-darwin     .             x86_64        .              .
+macos      aarch64-apple-darwin    .             aarch64       .              .
```
DilumAluthge, line 3:
```suggestion
macos      aarch64-apple-darwin    true          aarch64       .              .
```

DilumAluthge:
Why are the tests failing on Intel macOS?

gbaraldi:
I don't know, but both x86 and arm have thousands of segfaults so i would guess that has to do with it.

DilumAluthge:
@staticfloat It must be due to seatbelt, right?

DilumAluthge:
If I look at the raw log, before the many segfaults start, I see this:

```
Got exception outside of a @test

  LoadError: IOError: bind: name too long (ENAMETOOLONG)

  Stacktrace:

    [1] bind(server::Sockets.PipeServer, name::String)

      @ Sockets /Users/sabae/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-solstice.0/build/default-solstice-0/julialang/julia-buildkite/julia-a5fe945b4f/share/julia/stdlib/v1.9/Sockets/src/PipeServer.jl:59
```

I'm not sure if that's related to the segfaults, or if those are two separate issues.

gbaraldi:
I've seen those segfaults in the wild https://github.com/JuliaLang/julia/issues/44562, the socket error might be seatbelt however.

DilumAluthge:
Have you seen the segfaults on both Intel and Apple Silicon, or just on Apple Silicon?

gbaraldi:
I don't have an intel mac to test, but Ian had mentioned on the issue that he saw them on x86 as well.

gbaraldi:
@staticfloat in my last version I tried moving the keychain to `Library/Keychains` it didn't work I imagine becaue the agent overrides what home is. How would I make that work?

staticfloat:
We shouldn't need to move it to a different location.  Can you try using the full path to the keychain, even when passing to `unlock-keychain`?

DilumAluthge:
@staticfloat Why are there four different `Macos` groups in the Buildkite UI?

gbaraldi:
Oh yeah, macos doesn't have coreutils

staticfloat:
Because the job that launches the macos pipelines doesn't belong to the same group as the jobs its launching.  It's one of the more frustrating limitations of buildkite's jobs, see [this thread for more](https://forum.buildkite.community/t/group-merging-doesnt-work-as-expected/1858).

staticfloat:
@gbaraldi can you rebase on top of the latest `main` so you get the `git pager` fix?

staticfloat:
For debugging purposes, feel free to skip the tests by having the upload not depend on the tests. :P

pipelines/main/platforms/upload_macos.yml
```diff
@@ -0,0 +1,133 @@
...
+    # 1. The branch is `master`.
+    # 2. The branch is `release-*`.
+    # 3. The build is a tag build AND the tag is `v*`.
+    # 4. The pipeline is `julia-buildkite`.
+    if: (build.branch == "master") || (build.branch =~ /^release-/) || (build.tag =~ /^v/) || (pipeline.slug == "julia-buildkite")
+    depends_on:
+      # Wait for the builder to finish
+      - "build_${TRIPLET?}"
+      # Wait for the tester to finish
+      - "test_${TRIPLET?}"
```
DilumAluthge, line 14:
```suggestion
      # - "test_${TRIPLET?}" # TODO: uncomment this line
```

staticfloat:
I fixed the `ENAMETOOLONG`; the issue is that [libuv limits the length of unix domain socket paths](https://github.com/JuliaLang/libuv/blob/f872be3b6438b90b4690fe3ee7692c50bfd9c7c7/src/unix/pipe.c#L54-L55) to a value much smaller than the man page of `bind()` would indicate.  I've fixed this with https://github.com/JuliaCI/sandboxed-buildkite-agent/pull/9, allowing me to override the tempdir from a really long name to a much smaller prefix.

gbaraldi:
I wonder why `security -v list-keychains -s -d user "$$KEYCHAIN_PATH"` doesn't add the keychain to the list. It works locally. Maybe it tries to write somewhere that seatbelt doesn't allow and it fails silently?

DilumAluthge:
@staticfloat Is there a script or something that @gbaraldi can run on his local computer that will drop him into a Bash shell inside the seatbelt sandbox? That way he can reproduce the environment locally.

staticfloat:
> @staticfloat Is there a script or something that @gbaraldi can run on his local computer that will drop him into a Bash shell inside the seatbelt sandbox? 

Yes, [the `debug_shell.jl` script](https://github.com/JuliaCI/sandboxed-buildkite-agent/blob/main/macos-seatbelt/debug_shell.jl) will do that (note you'll need to setup a `config.toml` file first, just copy the example one over to `config.toml`).  I debugged this today though, and man, it was a pain.  I eventually broke down and turned off SIP on my Macbook Pro so that I could `dtruss` the `security` tool to figure out why it wasn't adding the keychain when it should have... and it turns out it was because we were overriding `HOME`, and the `${HOME}/Library/Preferences` directory did not exist.  Le sigh.

staticfloat:
Huzzah!  The codesigning is working!  Now to fix the `hdiutil` error and the distributed error.....

staticfloat:
Distributed failure should be fixed by https://github.com/JuliaLang/julia/pull/44799

gbaraldi:
You're a damn hero Elliot 🦸 . If only the command didn't fail silently, I will add a dmg upload step and I think we should be done. The DMG is also quarantined when downloaded. https://developer.apple.com/forums/thread/128166 this thread tells us to sign it too.

gbaraldi:
It seems to work now. It has the can't open terminal bug but that's on julia itself.

staticfloat:
The Pkg failure should be fixed by: https://github.com/JuliaLang/Pkg.jl/pull/3045

staticfloat:
Pkg failure identified and fix pending here: https://github.com/JuliaLang/Pkg.jl/pull/3050

staticfloat:
@gbaraldi I'm going to rebase this and start DRY'ing it a bit.  But I think we're getting about ready for a merge!

staticfloat:
Pkg problems are worked around as of https://github.com/JuliaLang/julia/pull/44828

So we can merge this when ready.

gbaraldi:
🥳 It's purple and green 🥳

staticfloat:
Nice work, Gabriel!

PR #51, opened 2022-03-10
Run a selected subset of tests under ASAN

tkf:
Applying https://github.com/JuliaLang/julia/pull/44546 here.


DilumAluthge (approved):
Assuming CI passes, this seems good to me.

DilumAluthge:
cc: @timholy

DilumAluthge:
We should probably increase the timeout?

It's unfortunate that it takes so long...

timholy:
If this isn't feasible, no sweat. I wondered if doing more than just build might catch some bugs. Of the CI failures, I'm not sure about many of them, but the ` LoadError: could not load library "libjulia-codegen"` might be an `so` filename issue?

tkf:
That's probably because the codgen tests are not written for `julia-debug`. There's only `libjulia-codegen-debug.so` etc. in this build.

DilumAluthge (approved):

PR #52, opened 2022-03-10
We must use `./` when referring to an external-buildkite filename

staticfloat:
This was documented, and summarily ignored by me as I added these plugin
stanzas.  We don't hit this on our own self-CI, because we override
these values.


staticfloat:
I've already updated the webui on `julia-master` and here.

DilumAluthge (approved):

PR #53, opened 2022-03-10
Try giving `rr` a maximum of 16 cores

staticfloat:
Let's just see what happens

### DIFF
--- a/utilities/rr/rr_capture.jl
+++ b/utilities/rr/rr_capture.jl
@@ -55,16 +55,18 @@ end
 
 @info "We will run the command under rr"
 
+const MAX_SUPPORTED_RR_NUM_CORES = 16
+
 const build_number                      = get_from_env("BUILDKITE_BUILD_NUMBER")
 const job_name                          = get_from_env("BUILDKITE_STEP_KEY")
 const commit_full                       = get_from_env("BUILDKITE_COMMIT")
 const commit_short                      = first(commit_full, 10)
 const JULIA_TEST_RR_TIMEOUT_MINUTES     = get(ENV,  "JULIA_TEST_RR_TIMEOUT_MINUTES", "120")
 const timeout_minutes                   = parse(Int, JULIA_TEST_RR_TIMEOUT_MINUTES)
-const JULIA_TEST_NUM_CORES              = get(ENV,  "JULIA_TEST_NUM_CORES", "16")
+const JULIA_TEST_NUM_CORES              = get(ENV,  "JULIA_TEST_NUM_CORES", "$(MAX_SUPPORTED_RR_NUM_CORES)")
 const julia_test_num_cores_int          = parse(Int, JULIA_TEST_NUM_CORES)
 const num_cores = min(
-    16,
+    MAX_SUPPORTED_RR_NUM_CORES,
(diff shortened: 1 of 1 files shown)


DilumAluthge (approved):
Let's give it a try!

utilities/rr/rr_capture.jl
```diff
@@ -55,16 +55,18 @@ end
...
+const MAX_SUPPORTED_RR_NUM_CORES = 16
+
 const build_number                      = get_from_env("BUILDKITE_BUILD_NUMBER")
 const job_name                          = get_from_env("BUILDKITE_STEP_KEY")
 const commit_full                       = get_from_env("BUILDKITE_COMMIT")
 const commit_short                      = first(commit_full, 10)
 const JULIA_TEST_RR_TIMEOUT_MINUTES     = get(ENV,  "JULIA_TEST_RR_TIMEOUT_MINUTES", "120")
 const timeout_minutes                   = parse(Int, JULIA_TEST_RR_TIMEOUT_MINUTES)
-const JULIA_TEST_NUM_CORES              = get(ENV,  "JULIA_TEST_NUM_CORES", "8")
+const JULIA_TEST_NUM_CORES              = get(ENV,  "JULIA_TEST_NUM_CORES", "$(MAX_SUPPORTED_RR_NUM_CORES)")
```
staticfloat, line 66:
```suggestion
const JULIA_TEST_NUM_CORES              = get(ENV,  "JULIA_TEST_NUM_CORES", "$(Sys.CPU_THREADS)")
```

DilumAluthge: 👍

staticfloat:
Interestingly, it doesn't seem to have helped runtime much.  Still at over 1 hour.  @Keno is there a fundamental limitation inside of `rr` that more than 8 cores doesn't help?

Keno:
The variable sets how many core rr pretends to julia that the machine has. It's for testing threads defaults and making sure that julia doesn't allocate unnecessary per-core buffers, which is slow under rr. It doesn't otherwise affect parallelism.

staticfloat:
When we change this value, Julia launches a different number of distributed workers when running the test suite.  Compare the worker numbers [here](https://buildkite.com/julialang/julia-buildkite/builds/158#07086aa0-6799-4848-a233-0c5bbe645891/351-427) with [here](https://buildkite.com/julialang/julia-buildkite/builds/153#5e1d3098-368e-4479-8825-0b4fe6f95b43/362-438).  Are you saying that even though Julia launches multiple distributed workers, the workers are serialized in their execution?

Keno:
No, the workers are not serialized. The number does set the maximum default for julia's reflection on the number of cores, so that can help. From that output, it looks like the Profile test is taking a large amount of time and limiting the overall end-to-end time. It does make sense that signal heavy code like Profile would slow down more under rr. I can look into if we can make that faster.

staticfloat:
If that's the case, let's be optimistic and use 16 threads here as well, in anticipation of some future `rr` magic improving things.

It would appear that moving from 8 to 16 cores has negatively impacted a good number of our tests.  Looks like, on average, most tests are reporting ~10% longer runtime.

PR #56, opened 2022-03-11
`upload_linux.yml`: only upload to S3 if the branch is `master` or `release-*` and it is not a pull request build

DilumAluthge:


DilumAluthge:
This should have failed the cryptic signature check, but it didn't.

staticfloat (changes requested):
We want to do this if it is master or release-*

staticfloat (approved):

PR #57, opened 2022-03-11
Signed jobs: make sure that all the necessary files are checked when verifying the signatures

DilumAluthge:


DilumAluthge:
Note: I haven't yet copied these changes into the web UI.

DilumAluthge:
This seems incorrect. We shouldn't have to modify the web UI this often.

Instead, the `launch_unsigned_jobs.yml` pipeline should itself use cryptic.

staticfloat:
Yeah agreed. We need to give cryptic enough flexibility that it can coexist with things like the arches files.

PR #61, opened 2022-03-12
Skip network-related tests when running under `rr`.

staticfloat:
It's a shame, but let's avoid ballooning our `rr` traces by recording
all the network activity as well.


DilumAluthge:
I'm not really a fan of this, especially if we can get around the file size limitation by just splitting it up into parts.

That being said, I guess I won't put up too much of a fuss 😂 .

As long as we are still running all of the tests (including all of the network tests) on the main (non-rr) `linux-x86_64-glibc` tester.

DilumAluthge (approved):

PR #62, opened 2022-03-12
Prevent multiple pipelines from uploading to S3 simultaneously

staticfloat:


DilumAluthge (approved):
Superficially, it looks good to me.

DilumAluthge:
In theory, we can interpolate the branch name into the name of the concurrency group. So e.g. in theory release-1.7 and release-1.8 could be different concurrency groups.

staticfloat:
That's true, but in practice I don't think it matters; the jobs should be pretty fast.

PR #63, opened 2022-03-13
Modify the existing `rr` tester job to use an assert build of Julia (instead of a regular build)

### LINKED ISSUE #40: TODO: add a pair of Buildkite jobs (e.g. `package_linux64_assert` and `tester_linux64_assert`) that builds an assert build of Julia and then runs the Julia test suite using the assert build

DilumAluthge:
Fixes #40

## Description

This PR:
1. Adds a new build job ("packager") that builds Julia with `FORCE_ASSERTIONS=1` and `LLVM_ASSERTIONS=1`.
2. Modifies the existing `rr` tester job so that it uses the assert build instead of a regular build.

## Timeouts

The combination of `rr` and an assert build is going to be very slow. Therefore, this PR increases the timeouts for the `rr` tests.

| Environment variable                     | Value   | Δ            |
| ---------------------------------------- | ------- | ------------ |
| `JULIA_TEST_RR_RUNTESTS_TIMEOUT_MINUTES` | 4 hours |              |
| `JULIA_TEST_RR_CLEANUP_TIMEOUT_MINUTES`  | 5 hours | + 60 minutes |
| `BUILDKITE_TIMEOUT`                      | 6 hours | + 60 minutes |

1. `JULIA_TEST_RR_RUNTESTS_TIMEOUT_MINUTES`: This is the amount of time that `Base.runtests("all")` gets. If this timeout is exceeded, we try to kill the `rr` process with `SIGTERM`.
2. `JULIA_TEST_RR_CLEANUP_TIMEOUT_MINUTES`: After we try to kill the `rr` process with `SIGTERM`, we wait some additional amount of time, corresponding to the `Δ` column in the `JULIA_TEST_RR_TIMEOUT_MINUTES_AFTER_CLEANUP` row. This period of time includes the time for `rr` to cleanup and exit; it also includes the time to upload the `rr` traces as Buildkite artifacts. If `rr` fails to cleanup and upload within this `Δ` amount of time, we kill the `rr` process with `SIGKILL`.
3. `BUILDKITE_TIMEOUT`: after we kill the `rr` process with `SIGKILL`, we should always see the job end immediately. So this additional `Δ` should never actually be used, unless something is going very wrong.

### DIFF
--- a/pipelines/main/platforms/build_linux.arches
+++ b/pipelines/main/platforms/build_linux.arches
@@ -1,11 +1,11 @@
-# OS       TRIPLET                 ALLOW_FAIL    ARCH_ROOTFS    MAKE_FLAGS     TIMEOUT       ROOTFS_TAG    ROOTFS_HASH
-# linux    i686-linux-gnu          .             i686           .              .             v4.8          b6dffc772ab4c2cd7fd4f83459308f6f0d89b957
-linux      x86_64-linux-gnu        .             x86_64         .              .             v5.1          f689ba6acab6c13a3da4ff62e3445dd61c3b5ab0
-# linux    aarch64-linux-gnu       .             aarch64        .              .             ----          ----------------------------------------
-# linux    armv7l-linux-gnueabihf  .             armv7l         .              .             ----          ----------------------------------------
-# linux    powerpc64le-linux-gnu   .             powerpc64le    .              .             ----          ----------------------------------------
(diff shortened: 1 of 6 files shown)


DilumAluthge:
@staticfloat Looks like the arches parsing script doesn't like something I did.

DilumAluthge:
@staticfloat Any idea what I'm doing wrong here?

DilumAluthge:
Ugh, is it the fact that I have two flags to pass to make, and they're space-separated?

DilumAluthge:
@staticfloat How can I make sure that the assert tester is actually using an assert build of Julia?

PR #68, opened 2022-03-17
S3 upload: only upload if the pipeline is `julia-master` or `julia-release-*`

DilumAluthge:
We have a variety of pipelines, including:

| Pipeline slug            | URL                                                    |
| ------------------------ | ------------------------------------------------------ |
| `julia-buildkite`        | https://buildkite.com/julialang/julia-buildkite        |
| `julia-master`           | https://buildkite.com/julialang/julia-master           |
| `julia-master-scheduled` | https://buildkite.com/julialang/julia-master-scheduled | 
| `julia-release-1-dot-6`  | https://buildkite.com/julialang/julia-release-1-dot-6  | 
| `julia-release-1-dot-7`  | https://buildkite.com/julialang/julia-release-1-dot-7  |
| `julia-release-1-dot-8`  | https://buildkite.com/julialang/julia-release-1-dot-8  |

We don't want all of those pipelines to be uploading to S3, only some of them. So this PR adds conditional statements based on the pipeline slug.

### DIFF
--- a/.buildkite/hooks/post-checkout
+++ b/.buildkite/hooks/post-checkout
@@ -22,3 +22,6 @@ fi
 if ! buildkite-agent meta-data exists BUILDKITE_JULIA_VERSION; then
     buildkite-agent meta-data set BUILDKITE_JULIA_VERSION "$(git rev-parse HEAD)"
 fi
+
+# Export S3_BUCKET_PREFIX, to force our CI to upload to a different prefix than the actual CI pipeline would.
+export S3_BUCKET_PREFIX="ephemeral/bin"
--- a/pipelines/main/platforms/upload_linux.yml
+++ b/pipelines/main/platforms/upload_linux.yml
@@ -1,10 +1,8 @@
 steps:
   - label: ":linux: upload ${TRIPLET?}"
     key:   "upload_${TRIPLET?}"
-    # We only upload to S3 if all of the following are true:
-    # 1. This is not a pull request build.
-    # 2. The branch is `master` or `release-*`.
-    if: (build.pull_request.id == null) && ((build.branch == "master") || (build.branch =~ /^release-/))
+    # We only upload to S3 if the branch is `master` or `release-*`.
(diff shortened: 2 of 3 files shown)


staticfloat:
Maybe we continue to allow this repository to upload, but it uploads to a different folder?

DilumAluthge:
FWIW, I don't think we're actually testing the upload step right now. Because it's skipped on PRs, and then the default branch of this repo is called `main` not `master`.

DilumAluthge:
> Maybe we continue to allow this repository to upload, but it uploads to a different folder?

This sounds like a good plan.

DilumAluthge:
What do you think about logic like this:

1. If the pipeline slug is `julia-master` or `julia-release-*`, then:
    - If (it is not a pull request build) AND (it is a push to `master` OR a push to `release-*` or a tag), then upload to the main S3 folder.
    - Else, do not upload.
2. If the pipeline slug is `julia-buildkite`, then upload to a special separate S3 folder that we clean out much more frequently (e.g. we delete objects in that folder after 3 days).
4. Else, do not upload.

DilumAluthge:
> Yeah, that's fine by me. I think we can probably make this easy on ourselves by adding a repository hook here that sets an environment variable like `S3_UPLOAD_BUCKET_NAME`, then we just use `$${S3_UPLOAD_BUCKET_NAME:-julialangnightlies}` in our upload job.

That sounds good to me! Can you make a PR?

staticfloat:
@DilumAluthge I've simplified this somewhat; first off, we upload if we're on `master` or `release-*`, no matter the pipeline.  This is to make the behavior a little more predictable, and to allow us to do things like push to `release-1.8` on this repository, and have things "just work", for our own testing.

DilumAluthge:
We'll still need to upload to different folders, to make sure that people downloading the 1.8-nightly don't accidentally download builds from this repo.

staticfloat:
Right, that is also included in this PR by setting environment variables in the post-checkout hook.

PR #70, opened 2022-03-17
Simplify the logic for "don't upload assert builds to S3"

DilumAluthge:
I think this is a better way to implement the "don't upload assert builds to S3" logic, because it uses the Buildkite `if` conditional.


DilumAluthge:
@staticfloat What do you think the best solution is here?

PR #71, opened 2022-03-17
Short-term fix to fix CI

DilumAluthge:


DilumAluthge:
@staticfloat The underlying issue here is that if the builder name and tester name don't match, then the upload job will hang indefinitely because its dependencies will never resolve, and thus the overall job will fail.

DilumAluthge:
So if you look at Julia master, e.g. https://buildkite.com/julialang/julia-master/builds/10154, and look at the `upload x86_64-linux-gnuassert` job: https://buildkite.com/julialang/julia-master/builds/10154#22f0c4fa-cfa1-4807-8dc7-1b572174184e

In the "timeline", it shows:

> Step Dependencies
> test_x86_64-linux-gnuassert
> build_x86_64-linux-gnuassert

But of course there is no job named `test_x86_64-linux-gnuassert`. (The tester was named `test_x86_64-linux-gnuassertrr`.) So the dependency is never satisfied.

staticfloat:
Let's just not upload the assert build at all.

DilumAluthge:
> Let's just not upload the assert build at all.

That's already the case: https://github.com/JuliaCI/julia-buildkite/blob/2d14144704f2b7c3327b73e631f8f24406f46f35/pipelines/main/platforms/upload_linux.yml#L43

I might have a better fix, though.

staticfloat:
Okay good, because I don't think this PR will work, because we can't duplicate keys.

PR #76, opened 2022-03-22
Upload tag builds to S3; build and upload musl to S3

DilumAluthge:


DilumAluthge:
```
upload failed: ./julia-95da0d8172-musl-x86_64.tar.gz to s3://julialangnightlies/ephemeral/bin/musl/x86_64/1.9/julia-95da0d8172-musl-x86_64.tar.gz An error occurred (AccessDenied) when calling the CreateMultipartUpload operation: Access Denied
```

@staticfloat Maybe you need to create the "ephemeral" folder?

staticfloat (approved):
I realize the tests are failing, but better than nothing.

DilumAluthge:
Yeah, I think we assume that musl tests will usually fail, which is why I marked it as an allowed failure.

DilumAluthge:
@staticfloat I think something has gone horribly wrong. The very first job seems to be hanging.

DilumAluthge:
@staticfloat Could this be related to the current GitHub downtime?

PR #79, opened 2022-03-24
Disable `git`'s internal pager when printing `git log -1`

staticfloat:
This has the unfortunate side-effect of freezing up our CI pipelines.


staticfloat:
@DilumAluthge This should fix the "random freezing" we have been seeing, such as in [this recent build](https://buildkite.com/julialang/julia-buildkite/builds/316#1da478f6-b8a4-46f3-b4f7-53a2044888c1).  The issue is that the commit message was so long that `git` automatically invoked a pager, which meant that it was waiting for standard input.  Easy fix; just disable the pager.

DilumAluthge:
If you like, we could also pipe to `head -n 20` or something like that.

PR #81, opened 2022-03-29
Fix a bug (caused by a typo) that resulted in the `rr` timeouts not being handled correctly

DilumAluthge:
Diagnosed in https://github.com/JuliaLang/julia/pull/44783

### DIFF
--- a/utilities/rr/rr_capture.jl
+++ b/utilities/rr/rr_capture.jl
@@ -170,7 +170,7 @@ mktempdir(temp_parent_dir) do dir
 
         # Start asynchronous timer that will kill `rr`
         @async begin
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@staticfloat Is there any way to print an message to stderr when an exception has been thrown inside an `@async` task, without needing to `wait` on that task?

staticfloat:
Starting in Julia v1.7+, we have [`errormonitor(::Task)`](https://github.com/JuliaLang/julia/pull/39518/files#diff-32f72b7e244562a3772085ca0dedcd4b717dd0be3535f73959b31e841f46d0aaR425).

PR #83, opened 2022-03-31
Fix another timeout-related typo in the `rr_capture.jl` script

DilumAluthge:


DilumAluthge:
See https://github.com/JuliaLang/julia/pull/44783 and https://github.com/JuliaLang/julia/pull/44783 for context.

PR #85, opened 2022-04-03
Manually create our own "pseudo-groups" using `depends_on` (instead of using the built-in Buildkite group functionality); also, add a GitHub commit status for each operating system

DilumAluthge:


staticfloat:
I think I see what you're trying to do here; use `depends_on` to create a single job for each OS..... but I feel it's actually a little confusing.  I think it's not worth it, and we should just stick with either no groups at all, or partial groups where it makes sense, and just hope that BK eventually fixes their group mechanic.

DilumAluthge:
I think it might actually be better to just go back to no groups. I think it's confusing that we have multiple "Linux" groups, multiple "macOS" groups, etc..

PR #87, opened 2022-04-08
Four Buildkite groups, four GitHub commit statuses: `build`, `check`, `test`, `upload`

DilumAluthge:


DilumAluthge:
@staticfloat The four different `upload` groups are not being merged. Is that a bug in the `cryptic` plugin?

The other three groups (`build`, `check`, and `test`) are being merged correctly.

DilumAluthge:
Here are the commands that I ran to re-sign all of the pipelines:

```
git clone git@github.com:JuliaCI/julia-buildkite.git
cd julia-buildkite
git checkout dpa/groups
git clone https://github.com/staticfloat/cryptic-buildkite-plugin.git
cd cryptic-buildkite-plugin
git checkout sf/group_capable
cd ..

# Copy the agent private key into the current directory
export AGENT_PRIVATE_KEY_PATH=$(pwd)/agent.key

make sign_treehashes

# Delete the agent private key file once we are done
rm -rf agent.key
```

PR #91, opened 2022-04-08
Add one new Buildkite group (`Allow Fail`), but do not add a GitHub commit status for that group

### LINKED ISSUE #47: "Allowed to fail" jobs should go into their own groups
This is what our Buildkite groups currently look like:

I'm not thrilled about the fact that the `Linux` group is red. It's red (with a warning sign) because one of the jobs in the `Linux` group failed but has `soft-fail: true`. Specifically, the `asan` job failed but has `soft-fail: true`.

Now, this particular example is not a good example, because on Julia master we can actually take off the soft-fail for `asan`, since `asan` is passing on Julia master. But, in general, we will always have jobs that we want to run but allow to fail. And I don't think those should "pollute" our green groups.

I would like us to put "allowed to fail" jobs into their own groups, and explicitly put the text `allowed to fail` in the name of the group so that user can easily tell that it's okay for those groups to be red. We have two choices:

1. Platform-specific "allowed to fail" groups for each platform.
2. A single "allowed to fail" group for all platforms.

Here is what option 1 would look like:
1. Infrastructure
2. Windows
3. Linux
4. macOS
5. FreeBSD
6. Allowed to fail (Windows)
7. Allowed to fail (Linux)
8. Allowed to fail (macOS)
9. Allowed to fail (FreeBSD)

Here is what option 2 would look like:
1. Infrastructure
2. Windows
3. Linux
4. macOS
5. FreeBSD
6. Allowed to fail

DilumAluthge:
This pull request adds one new Buildkite group:
1. `Allow Fail`

However, if desired, I can instead split that group up into two separate groups as such:
1. `Allow Fail (Build)`
2. `Allow Fail (Test)`

Fixes #47

### DIFF
--- a/pipelines/main/platforms/build_linux.arches
+++ b/pipelines/main/platforms/build_linux.arches
@@ -3,12 +3,12 @@
 linux      x86_64-linux-gnu          .             x86_64         x86_64         CFLAGS=-Werror,CXXFLAGS=-Werror                                         .             v5.8          582b5d44bbc24d2a58ab609c5a520b3fd102e504
 linux      x86_64-linux-gnuassert    .             x86_64         x86_64         FORCE_ASSERTIONS=1,LLVM_ASSERTIONS=1,CFLAGS=-Werror,CXXFLAGS=-Werror    .             v5.8          582b5d44bbc24d2a58ab609c5a520b3fd102e504
 # linux    aarch64-linux-gnu         .             aarch64        aarch64        .                                                                       .             ----          ----------------------------------------
-# linux    armv7l-linux-gnueabihf    .             armv7l         armv7l         .                                                                       .             ----          ----------------------------------------
-# linux    powerpc64le-linux-gnu     .             powerpc64le    powerpc64le    .                                                                       .             ----          ----------------------------------------
 musl       x86_64-linux-musl         .             x86_64         x86_64         .                                                                       .             v5.8          dc23dbeb02f9b85c8c8c7991ee8cfcae8f4c809b
 
 # These special lines allow us to embed default values for the columns above.
 # Any column without a default mapping here will simply substitute a `.` to the empty string
+
+#default GROUP build
 
 # Of course we do not allow jobs to fail typically
 #default ALLOW_FAIL false
--- a/pipelines/main/platforms/build_linux.arches
+++ b/pipelines/main/platforms/build_linux.arches
@@ -1,17 +1,12 @@
-# OS       TRIPLET                   ALLOW_FAIL    ARCH           ARCH_ROOTFS    MAKE_FLAGS                              TIMEOUT       ROOTFS_TAG    ROOTFS_HASH
-# linux    i686-linux-gnu            .             x86_64         i686           .                                                                       .             ----          ----------------------------------------
-linux      x86_64-linux-gnu          .             x86_64         x86_64         CFLAGS=-Werror,CXXFLAGS=-Werror                                         .             v5.8          582b5d44bbc24d2a58ab609c5a520b3fd102e504
-linux      x86_64-linux-gnuassert    .             x86_64         x86_64         FORCE_ASSERTIONS=1,LLVM_ASSERTIONS=1,CFLAGS=-Werror,CXXFLAGS=-Werror    .             v5.8          582b5d44bbc24d2a58ab609c5a520b3fd102e504
-# linux    aarch64-linux-gnu         .             aarch64        aarch64        .                                                                       .             ----          ----------------------------------------
-musl       x86_64-linux-musl         .             x86_64         x86_64         .                                                                       .             v5.8          dc23dbeb02f9b85c8c8c7991ee8cfcae8f4c809b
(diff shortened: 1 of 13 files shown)


pipelines/main/platforms/build_linux.arches
```diff
@@ -3,13 +3,13 @@
...
 # linux    aarch64-linux-gnu         .             aarch64        aarch64        .                                                                       .             ----          ----------------------------------------
-# linux    armv7l-linux-gnueabihf    .             armv7l         armv7l         .                                                                       .             ----          ----------------------------------------
-# linux    powerpc64le-linux-gnu     .             powerpc64le    powerpc64le    .                                                                       .             ----          ----------------------------------------
 musl       x86_64-linux-musl         .             x86_64         x86_64         .                                                                       .             v5.8          dc23dbeb02f9b85c8c8c7991ee8cfcae8f4c809b
 
 # These special lines allow us to embed default values for the columns above.
 # Any column without a default mapping here will simply substitute a `.` to the empty string
 
+#default GROUP build
+
```
staticfloat, line 12: I don't think you're actually using this?

pipelines/main/platforms/build_linux.arches
```diff
@@ -3,13 +3,13 @@
...
 # Of course we do not allow jobs to fail typically
 #default ALLOW_FAIL false
```
staticfloat, line 14: Now that we have the `soft_fail.arches`, we don't need this anymore, right?

pipelines/main/platforms/build_linux.arches
```diff
@@ -3,13 +3,13 @@
 linux      x86_64-linux-gnu          .             x86_64         x86_64         CFLAGS=-Werror,CXXFLAGS=-Werror                                         .             v5.8          582b5d44bbc24d2a58ab609c5a520b3fd102e504
 linux      x86_64-linux-gnuassert    .             x86_64         x86_64         FORCE_ASSERTIONS=1,LLVM_ASSERTIONS=1,CFLAGS=-Werror,CXXFLAGS=-Werror    .             v5.8          582b5d44bbc24d2a58ab609c5a520b3fd102e504
 # linux    aarch64-linux-gnu         .             aarch64        aarch64        .                                                                       .             ----          ----------------------------------------
```
staticfloat, line 5: Did this line survive for a reason?

pipelines/main/platforms/test_macos.soft_fail.arches
```diff
@@ -0,0 +1,11 @@
+# OS       TRIPLET                 ALLOW_FAIL    ARCH          MAKE_FLAGS     TIMEOUT
+macos      aarch64-apple-darwin    true          aarch64       .              .
+
+# These special lines allow us to embed default values for the columns above.
+# Any column without a default mapping here will simply substitute a `.` to the empty string
+
+# Of course we do not allow jobs to fail typically
+#default ALLOW_FAIL false
```
staticfloat, line 8: Shouldn't we just drop `ALLOW_FAIL` completely and just set it to `true` when launching all `soft_fail.arches` files?

pipelines/main/platforms/build_linux.arches
DilumAluthge: Yes. I wanted to ask you - do we expect that building `aarch64-linux-gnu` will always succeed?

I put the tester for `aarch64-linux-gnu` in the "allowed to fail" list, but I wasn't sure about the builder.

pipelines/main/platforms/build_linux.arches
DilumAluthge: Correct. I've removed it.

pipelines/main/platforms/build_linux.arches
DilumAluthge: Correct. I've removed it.

pipelines/main/platforms/test_macos.soft_fail.arches
DilumAluthge: Yes, that is much easier. I've implemented that change.

DilumAluthge:
@staticfloat What is your opinion on one group (`Allow Fail`) versus two groups (`Allow Fail (Build)` and `Allow Fail (Test)`).

I went with one group because I wanted to minimize the total number of groups that we have. But I don't feel particularly strongly either way, and I would be fine with switching to two groups.

staticfloat:
I'm happy with one group.  Now that I'm looking at this though, I kind of wish we could filter out the `ALLOW_FAIL` builds from a single `.arches` file so we don't have to split into multiple files.  But I don't care that much.  :P

pipelines/main/platforms/build_linux.arches
DilumAluthge: Cool, so let's plan to keep the builder as "normal", and the tester as "allowed to fail"?

DilumAluthge:
> I kind of wish we could filter out the `ALLOW_FAIL` builds from a single `.arches` file so we don't have to split into multiple files

I think that would be really convenient, and I'd like to have that.

I doubt I'll get around to implementing that anytime soon, but the nice thing is we can implement that optimization behind the scenes without making any user-facing changes.

PR #92, opened 2022-04-09
[DO NOT MERGE] Split the single `Allow Fail` group into two groups: `Allow Fail (Build)` and `Allow Fail (Test)`

DilumAluthge:
Do not merge this pull request.

This is really just a note to myself. If, in the future, we decide we want to split the `Allow Fail` group into multiple groups, this is a demonstration of how we would do it.

### DIFF
--- a/pipelines/main/launch_unsigned_builders.yml
+++ b/pipelines/main/launch_unsigned_builders.yml
@@ -81,7 +81,7 @@ steps:
               .buildkite/pipelines/main/platforms/test_macos.yml
         agents:
           queue: "julia"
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Honestly, it doesn't look too bad. If we do decide to switch to two groups in the future, I don't think it would be terrible. But let's stick with the single group for now, in the interest of minimizing the total number of Buildkite groups.

---

Here's how it looks in GitHub (note that we don't show any of the "allow fail" commit statuses in GitHub):

github

And here's how the Buildkite groups look in the Buildkite UI:

buildkite

PR #93, opened 2022-04-09
Only run the Apple Silicon tester if the branch is `master` or `main`

DilumAluthge:


staticfloat:
Why don't we try just running it on `master` for now?  That should cut down the testing burden significantly.  I still want to see the results of the tests.

DilumAluthge:
Ugh that's not right.... we only want to apply this change to the Apple Silicon tester; we don't want to apply it to the x86_64 tester.

DilumAluthge:
I was hoping to use Buildkite conditionals, but I don't think we can do that here, since we also need to condition on the arch.

DilumAluthge:
@staticfloat Is there any way to debug Buildkite conditionals? I'd like to print what the value of `build.env("ARCH")` actually is.

staticfloat:
It looks to me like you do need quotes: https://buildkite.com/docs/pipelines/conditionals#example-expressions

DilumAluthge:
Yeah, that's what I had thought. Buildkite support told me in email to try it without the quotes. I just replied to their email with a link to the latest log.

DilumAluthge:
I CC'd you on my reply.

PR #95, opened 2022-04-11
Build for macOS with `-Werror`

giordano:
With https://github.com/JuliaLang/julia/pull/44807 merged, we should now be able to build for macOS with `-Werror`, which is useful since here we use Clang, which tends to be stricter than GCC.


giordano:
All build jobs are successful, failures in test are unrelated :slightly_smiling_face:

DilumAluthge (approved):

PR #98, opened 2022-04-13
[DO NOT MERGE] Out of curiosity, let's see how big the rr trace file is with all of the tests

DilumAluthge:


DilumAluthge:
This will also serve as a test that my `split` code is working correctly.

DilumAluthge:
Alright, so two files, with sizes 4.56 GB and 618 MB, for a total of 5.178 GB. Not terrible.

But since the Buildkite limit is 5 GB per individual file, we can't upload it as a single file, which is annoying. So, for now, let's stick with the approach in https://github.com/JuliaCI/julia-buildkite/pull/99

DilumAluthge:
> This will also serve as a test that my `split` code is working correctly.

Also, yes, this confirmed that the `split` code (for splitting a single large file into multiple smaller files) is working correctly.

PR #102, opened 2022-04-17
Get rid of the "scheduled" pipeline, and move all of its jobs to the "main" pipeline

### LINKED ISSUE #54: TODO: re-enable the daily scheduled pipeline
https://buildkite.com/julialang/julia-master-scheduled

It has two jobs:
1. Code coverage
2. `USE_BINARYBUILDER=0`

### LINKED ISSUE #106: TODO: in the `coverage` job, skip the Pkg tests
With code coverage enabled and native code from the sysimage disabled, the Pkg tests take more than an hour to run. I think we can exclude the Pkg tests from the `coverage` job without losing too much `Base` code coverage.

DilumAluthge:
It turns out that it's real annoying to have to maintain two completely separate pipelines. So let's combine them all into a single pipeline, and just use Buildkite conditionals (the lines that start with `if:`) to distinguish between the three following disjoint categories of jobs:
1. Jobs that should only be run if the build is a `schedule` build. Currently this includes the following jobs:
    - The "from source" (`USE_BINARYBUILDER=0`) build job
    - The "from source" (`USE_BINARYBUILDER=0`) test job
    - The code coverage job
3. Jobs that should only be run if the build is NOT a `schedule` build. This is most of our jobs.
5. Jobs that should always be run. Currently, we have no such jobs.

Fixes #54 
Fixes #106


DilumAluthge:
I'll re-generate the signatures later.

DilumAluthge:
```
/cache/build/default-amdci5-5/julialang/julia-buildkite/Make.inc:1176: *** "Attempting to build OpenBLAS or SuiteSparse without a functioning fortran compiler!".  Stop.
```

@staticfloat Can you put `gfortran` in the `package_linux.x86_64` rootfs image? I need it for the "from source" (`USE_BINARYBUILDER=0`) job that we run on `master` on a once-daily schedule.

DilumAluthge:
If that's going to be a heavy lift, I can just switch to using the `llvmpasses.x86_64` rootfs image for the "from source" build; we don't upload those binaries to S3, so it doesn't matter if they use a very recent glibc.

DilumAluthge:
@staticfloat I went with the "use the `llvm_passes` image" approach for now, just to get this PR working. Let me know what you'd like to do long-term.

PR #103, opened 2022-04-17
Add documentation for how maintainers can re-generate all of the signatures

DilumAluthge:


DilumAluthge:
I wrote this documentation for the use case where you have the agent private key file, and you want to re-generate all signatures.

But we should also have the documentation for the (more common) use case where you only have the repository private key file, and you want to re-generate all signatures.

@staticfloat Could you add that documentation to this PR? You can push directly to my branch, or if you want you can post the instructions in a comment here, and I can add them. I'm guessing the difference will just be that the user won't set the `AGENT_PRIVATE_KEY_PATH` environment variable and will instead set a different environment variable that points to the path of the repository private key file?

PR #105, opened 2022-04-17
S3 upload: when printing the S3 URLs, print them in a format that the user can copy and paste into their browser

DilumAluthge:


DilumAluthge:
@staticfloat Since the path style is deprecated, how should we modify this path to instead print out the virtual-hosted-style?

PR #109, opened 2022-04-21
Put the "from source" (`USE_BINARYBUILDER=0`) jobs into their own `Source` Buildkite group

DilumAluthge:
Previously, the "from source" build job was in the regular `Build` Buildkite group, and the "from source" test job was in the regular `Test` Buildkite group. And we used Buildkite conditionals to skip those jobs on PRs, since we only want to run those jobs on the once-daily scheduled build.

But this caused a problem. Recall that the `Build` and `Test` Buildkite groups correspond to GitHub commit statuses that we display on pull requests. For some reason, if one or more jobs in a Buildkite group are skipped per the Buildkite conditiional, Buildkite will always display that Buildkite group's corresponding GitHub commit status as pending (yellow). So it will never turn green even once all the non-skipped jobs have completed. As a result, now, on PRs, the `Build` and `Test` commit statuses always stay yellow and never turn green.

I consider this to be a bug in Buildkite. But until they fix it, we will work around it by moving the "from source" jobs into their own `Source` Buildkite group.

### DIFF
--- a/pipelines/main/launch_unsigned_jobs.yml
+++ b/pipelines/main/launch_unsigned_jobs.yml
@@ -30,11 +30,6 @@ steps:
               bash .buildkite/utilities/arches_pipeline_upload.sh \
               .buildkite/pipelines/main/platforms/build_linux.arches \
               .buildkite/pipelines/main/platforms/build_linux.yml
-          GROUP="Build" \
-              ALLOW_FAIL="false" \
-              bash .buildkite/utilities/arches_pipeline_upload.sh \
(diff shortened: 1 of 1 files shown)


DilumAluthge:
My long-term plan is to have Buildkite fix their bug. Namely, skipped jobs in a Buildkite group should not impact the Buildkite group's corresponding GitHub commit status.

My medium-term plan is to switch away from using Buildkite conditionals and instead use regular Bash conditionals in our launch scripts.

PR #111, opened 2022-04-21
Add the build and test jobs for `i686-linux-gnu` and `aarch64-linux-gnu`

staticfloat:
My GCC toolchain script is finally bearing fruit!


DilumAluthge (approved):

staticfloat:
```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.0-DEV.426 (2022-04-21)
 _/ |\__'_|_|_|\__'_|  |  Commit 4dfef57535 (0 days old master)
|__/                   |

julia> Base.BinaryPlatforms.HostPlatform()
Linux i686 {cxxstring_abi=cxx11, julia_version=1.9.0, libc=glibc, libgfortran_version=5.0.0}
```
:tada:

DilumAluthge:
@staticfloat @Keno I'm seeing some OOM failures on the Linux i686 tester. Could those be related to the GC heuristics? I thought we had fixed those.

DilumAluthge:
@staticfloat I've retried the Linux i686 tester multiple times, but it is failing every time.

@Keno Maybe it's time for us to build an i686 (32-bit) version of `rr` in Yggdrasil? 😉

DilumAluthge:
I'm still seeming some test failures, and at least one of them is an OOM.

Do we need to further decrease the max RSS limit? Or should we maybe decrease the number of test workers from 16 down to something smaller, such as 14 or 12?

vtjnash:
Why are you limiting the usefulness of `--output-sync` in this PR?

DilumAluthge:
> Why are you limiting the usefulness of `--output-sync` in this PR?

It was an attempt to get the Buildkite timestamps to work. (Go to the log of one of the Linux jobs, and click "show timestamps".)

But even with `--output-sync=line`, the timestamps don't seem to be correct. Maybe Julia is buffering the output?

staticfloat:
I initially tried limiting worker RSS to something relatively low (so workers would get restarted if they started a run with a max RSS of 1.5 GB) but we're still getting the errors.

Looking at [this log](https://buildkite.com/julialang/julia-buildkite/builds/521#c4211684-ce0f-4cc6-ad76-8d2554861c91/416-874) we die in the `Pkg` tests due to `std::bad_alloc` inside of LLVM.  Looking at `dmesg`, there are no OOM killer declarations, so I figured this must be due to address space exhaustion.  Loading up a 32-bit userspace on the same machine (so same kernel), I can run the following:

```julia
xs = []
for idx in 1:1000
    push!(xs, zeros(UInt8, 10*1024^2))
    @info("bump!", idx, RSS=Int64(Sys.maxrss())/1024^2)
end
```
which eventually dies at:
```
┌ Info: bump!
│   idx = 353
└   RSS = 3719.48828125
ERROR: OutOfMemoryError()
Stacktrace:
 [1] Array
   @ ./boot.jl:461 [inlined]
 [2] Array
   @ ./boot.jl:470 [inlined]
 [3] zeros
   @ ./array.jl:587 [inlined]
 [4] zeros
   @ ./array.jl:583 [inlined]
 [5] top-level scope
   @ ./REPL[2]:2
```

So it looks like, theoretically at least, we should be able to allocate ~3.7GB of memory within a single process on these machines.

To get an idea of how much memory the `Pkg` tests require (the ones that died in that log), I ran the `Pkg` tests, with a loop reading from `/proc/$pid/statm` in another terminal:
```
while [ true ]; do cat /proc/${julia_pid}/statm; sleep 1; done
308726 168602 36310 2 0 224158 0
308019 168424 36310 2 0 223451 0
...
```

Taking the maximum of these values gives an estimate of how much memory just the Pkg tests can demand.  In this test, I saw `size` grow from `134609` to `598387`, and `resident` grow from `31690` to `444191`. which (with 4K pages) represents an increase of 1.8 and 1.6GB, respectively.  This (should) mean that starting a run with an RSS of 1.5GB should end up with us at ~3.1GB, or so I figure, which is still well below the 3.7GB limit I observed in the other test.

@vtjnash do you have any debugging tips here?  I'd like to understand what we can do to fix this, because as soon as I try and start running tests on an armv7l board, we're going to have much worse issues, as there we will legitimately only have 4GB of RAM total.  :P

vtjnash:
3.7 GB is somewhere around the theoretical limit for measurements given the physical limitations, but the practical limit is likely somewhere a bit less than 2 GB given other considerations. Pkg should not need to do anything that takes more than a hundred MB though.

staticfloat:
> but the practical limit is likely somewhere a bit less than 2 GB given other considerations

Okay; looking at the latest run, we see that `Pkg`, running on worker 18, jumped from [472.46](https://buildkite.com/julialang/julia-buildkite/builds/532#7e099947-2f57-4ed3-8117-38fcb4df7c26/433-807) to [1956.40](https://buildkite.com/julialang/julia-buildkite/builds/532#7e099947-2f57-4ed3-8117-38fcb4df7c26/433-892) MB RSS, which points to an RSS increase of ~1.5GB, which is about on par with what I measured earlier.  Assuming that we actually can't go higher than 2GB, that means that, to be safe, we would basically need to set the RSS limit to 500MB, which means that for all but the simplest tests we'd be restarting the worker immediately afterward.

> Pkg should not need to do anything that takes more than a hundred MB though.

I would be overjoyed to embed some memory debugging tools into our CI systems; I think it would be really interesting if we can have a worker dump memory statistics [just before getting killed](https://github.com/JuliaLang/julia/blob/fbec395fd049948618442569aa3a7708ceed2013/test/runtests.jl#L289) by the test suite RSS limit.  I'd love to dump things like total RSS, generated code size, memory belonging to Julia-reachable objects, unaccountable memory, etc...

staticfloat:
Also, we seem to have triggered the "timeout at the end of the testing section" bug on i686-linux-gnu [here](https://buildkite.com/julialang/julia-buildkite/builds/532#7e099947-2f57-4ed3-8117-38fcb4df7c26/433-896).  Exciting

staticfloat:
Profiling on aarch64 seems to be broken: https://buildkite.com/julialang/julia-buildkite/builds/532#4f193e60-7596-4547-8770-fe803022e9cd/429-905

DilumAluthge:
> we would basically need to set the RSS limit to 500MB, which means that for all but the simplest tests we'd be restarting the worker immediately afterward.

Is that a huge problem? It doesn't take terribly long to start a new worker.

DilumAluthge:
Alright, well I tried setting it to 500, but the job hangs and eventually hits the 90 minute timeout; I retried it a couple times but it timed out each time. So I guess setting it to 500 isn't a solution either.

PR #115, opened 2022-04-23
Upload linux jobs should only run on x86_64

staticfloat:


DilumAluthge:
This PR did not update the signature files, which caused this failure: https://buildkite.com/julialang/julia-master/builds/11332

I've updated the signature files in https://github.com/JuliaCI/julia-buildkite/pull/116

PR #118, opened 2022-04-26
Add `armv7l-linux-gnueabihf` builder, tester and uploader

staticfloat:
Because testing on armv7l is painful, we run only a very small set of
tests, pending greater investment in armv7l infrastructure.


staticfloat:
This is going to sit and wait for us to actually have armv7l builders again.

inkydragon:
Maybe we could get some free ARM computing resources with funding provided by Arm Inc.
(miniNodes/Open Source Lab/Equinix)
https://www.arm.com/solutions/infrastructure/works-on-arm

staticfloat:
And after many moons of inactivity, we have finally procured armv7l hardware!

Now begins the great debugging spree:

```
error: 'bt_context_t' {aka 'struct unw_tdep_context'} has no member named 'uc_mcontext'
```

giordano:
https://github.com/JuliaLang/julia/issues/47345

topolarity:
Would it be easier to manage the workers if we used Aarch64 machines running in 32-bit mode (assuming that we have those available)?

I did something like that recently to test out some ARMv7 changes, and it was surprisingly smooth with multi-arch on Debian.

giordano:
I'm not entirely sure, but I seem to remember https://github.com/JuliaLang/julia/issues/41064 couldn't be reproduced on aarch64 running the armv7l binary.

PR #123, opened 2022-05-26
Catch background task errors, copy mac64 binaries

### LINKED ISSUE #122: macOS nightlies changed URL, and out-of-date due to missing python
The macOS nightlies as linked on the home page and used by GitHub actions like setup-julia, are quite a bit out of date:

```
$ wget https://julialangnightlies-s3.julialang.org/bin/mac/x64/julia-latest-mac64.dmg
# mount dmg
$ ./julia               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.0-DEV.439 (2022-04-26)
 _/ |\__'_|_|_|\__'_|  |  Commit 7ad0e3deae (29 days old master)
|__/                   |
```

This is causing some issues with packages that assume latest master, such as GPUCompiler.jl.

Looking at the logs of the `upload x86_64-apple-darwin` job on Buildkite, https://buildkite.com/julialang/julia-master/builds/12357#0180fb98-58d5-4c98-bc3b-9dc178cf97c9, it seems like the binaries are now being uploaded to a different endpoint: s3://julialangnightlies/bin/macos/x64/julia-latest-macos64.dmg. However, the HTTP version of that, http://julialangnightlies-s3.julialang.org/bin/macos/x64/julia-latest-macos64.dmg, is not usable. The reason behind that is that the bucked is empty, because the upload failed:

```
$ s3cmd ls s3://julialangnightlies/bin/macos/x64/
                          DIR  s3://julialangnightlies/bin/macos/x64/1.9/
2022-05-25 15:32          833  s3://julialangnightlies/bin/macos/x64/julia-latest-macos64.tar.gz.asc
```

```
> Upload primary products to S3

.buildkite/utilities/upload_julia.sh:  /usr/local/bin/aws: /usr/local/Cellar/awscli/2.5.2/libexec/bin/python3.9: bad interpreter: 
No such file or directory
```

I'd argue that the exit code of the `aws` utility needs to be checked. That currently isn't the case, despite the `set -e`, because the `aws` invocations are backgrounded: https://github.com/JuliaCI/julia-buildkite/blob/0387123a0e8ff6d6db402c1bb96165be2c6d4e74/utilities/upload_julia.sh#L66=

**In addition**, I think it would be better if the endpoint wouldn't have changed. Unless there's a good reason to, this seems like a needlessly-breaking change that requires updating uses of the old URL (the homepage, the setup-julia GitHub action, etc).

staticfloat:
Backgrounded tasks don't fail execution with `set -e` unless they are explicitly `wait`'ed upon.

Fixes #122


maleadt:
Awesome, thanks!

So I take it you prefer to keep using the new URL scheme?

staticfloat:
Looking good: https://buildkite.com/julialang/julia-buildkite/builds/551#01810224-3ce2-41c7-96ad-8547c5c0ceaf/906-919

I'm not super happy with the old scheme as it's quite inconsistent.  I don't want to commit to a "new" URL scheme yet which is why I'm duplicating files all over the place right now, but eventually I will want to move to a new, more consistent, scheme.

For example, there's no reason to have the arch folder name `x64`, and the OS folder name be `macos`, but the filename say `-mac64`.  Instead, we should use the same terms everywhere, e.g. `/x86_64/macos/julia-latest-mac-x86_64.dmg`.

PR #126, opened 2022-05-29
Add a GitHub Actions CI job to this repo that verifies all of the signatures

### LINKED ISSUE #18: Add a CI job to this repo that just verifies all of the signatures?
One of the most frequent mistakes that I make is forgetting to update the signatures. Obviously we'll catch that when we run the test job on our test repo. But I would much rather catch that mistake sooner, before I merge the PR here.

What I'm envisioning is a very quick GitHub Action on this repo that simply takes in the repo public key and verifies all of the signatures against the public key.

DilumAluthge:
Fixes #18


DilumAluthge:
@staticfloat:
1. Where can I find the repository public key for the `JuliaLang/julia` repository?
2. Once I have the repository public key, what environment variable do I need to set so that `make verify_treehashes` can find the repository public key?

PR #128, opened 2022-05-29
Remove some uses of Buildkite conditionals and replace them with Bash conditionals

DilumAluthge:
Currently, on the once-daily scheduled jobs, the `Build`, `Check`, and `Test` GitHub commit statuses stay yellow and never turn green. This PR should fix that.


DilumAluthge:
Frustrating - it still creates a pending (yellow) commit status for e.g. `Build` even if the entire `Build` group is skipped. I'll have to try something else.

PR #130, opened 2022-06-06
On Linux i686, set the max rss to 500MB

DilumAluthge:


staticfloat:
We will also likely need to bump up the time limit to something like 4 hours.

staticfloat:
Wait, I take it back.  It passed in 45 minutes?!  I must have been seeing some other bug when it timed out after 2.5 hours so many times.

staticfloat (approved):

DilumAluthge:
:party_parrot:

DilumAluthge:
> I must have been seeing some other bug when it timed out after 2.5 hours so many times.

Yeah, I think that's a separate bug. I've been seeing it on Linux i686, but also on macOS x86_64. I don't think it's a timeout, I think it's an indefinite hang.

PR #132, opened 2022-06-07
Automatically collect core dumps on linux when testing

staticfloat:
When running Julia tests, segfaults can be difficult to find and
reproduce.  This adds automatic collection, inspection. compression and
uploading of core dump files to all linux testers.  It works by enabling
core dumps through `ulimit`, relying on our sandbox workers setting up a
good default core dump pattern [0], and then compressing the core dumps
with `zstd` before uploading to buildkite's artifact storage.  The
compression requires a tester rootfs that contains the `zstd` binaries,
which are available since rootfs release v5.17 [1].

We also stick in a few automatic `gdb` commands to give common debugging
output to allow developers to rapidly check common requests within the
web UI of buildkite itself.

[0] https://github.com/JuliaCI/sandboxed-buildkite-agent/commit/d64cacf719c5e20f4fc9abd50264459e0f4aefdf#diff-bdb8ed73fa1e3069f9f4eb7bf82cd431b578a6f462897ee8f73a4f6e3aacba9aR5
[1] https://github.com/JuliaCI/rootfs-images/commit/bdaf127c9b3bdecdbf7799dc2cfca2384bea5166


staticfloat:
This PR is being tested here: https://github.com/JuliaLang/julia/pull/45611

DilumAluthge (approved):

staticfloat:
Alright, this should be good to go, from my perspective.

PR #138, opened 2022-06-13
Disable tests with ASAN

tkf:
Implementing https://github.com/JuliaLang/julia/pull/45195#issuecomment-1153321913


tkf:
@DilumAluthge @vchuravy Do we want to keep running the tests? Since we know they fail anyway, maybe it's better to just remove it?

tkf:
Hmm... The step was waiting for an agent (until the build was canceled by a push). Is there something wrong with the patch?

> image
>
> --- https://buildkite.com/julialang/julia-buildkite/builds/615#01815bf3-6771-441a-bee6-2669b86e6bd8

PR #142, opened 2022-06-16
Add coredump uploading for macOS

staticfloat:
Also, prefer using `lldb` over `gdb` on all platforms, as it's more efficient.


staticfloat:
I will note that macOS produces _huge_ coredumps.  The smallest I've seen is 3.5GB in size (zstd compresses it down to ~800MB).  I can really see why people use minidumps.  Perhaps I will eventually integrate a minidump watcher/writer, but until then, we make use of buildkite's generous S3 hosting!

PR #146, opened 2022-06-21
Copy aarch64 mac binaries to the "old" naming style as well

### LINKED ISSUE #145: macOS aarch64 binaries not uploaded to old naming scheme
This was fixed for x64 in https://github.com/JuliaCI/julia-buildkite/pull/123/commits/2b1525e2fcd33f42c6ab3ba64c5582dbd00612a6 by @staticfloat, and the same should probably be done for aarch64 so that we can use:

```
https://julialangnightlies-s3.julialang.org/bin/mac/aarch64/julia-latest-macaarch64.tar.gz
```

Or at least that's the path that our Buildkite plugin expects: https://github.com/JuliaCI/julia-buildkite-plugin/blob/main/hooks/pre-command#L147-L150=. We could update the plugin to match the current path instead, but seeing how that's apparently not final yet maybe it's easier to duplicate the x64 hack?

staticfloat:
Fixes #145


maleadt:
Awesome, thanks Elliot!

PR #150, opened 2022-06-24
Upload test results if they are available

vchuravy:
Companion to https://github.com/JuliaLang/julia/pull/45799


pipelines/main/platforms/test_macos.yml
```diff
@@ -16,6 +16,9 @@ steps:
           - JuliaCI/external-buildkite#v1:
               version: "./.buildkite-external-version"
               repo_url: "https://github.com/JuliaCI/julia-buildkite"
+          - test-collector#v1.0.0:
+            files: "test/results.json"
+            format: "json"
```
vchuravy, line 21:
```suggestion
              files: "test/results.json"
              format: "json"
```

staticfloat (approved):

staticfloat (commented):
I can get us a `BUILDKITE_ANALYTICS_TOKEN`.  However, pursuant to our previous discussions about security and tokens, we have a few options:

1) Do the typical song and dance, but since we don't want to sign the entire test suite, we'd basically be allowing anyone to print out the test suite token.  I don't see this as that big of a problem, since the worst that can happen is that someone screws up our test suite results.  Annoying, but not fatal.

2) separate the test suite from the results uploading; basically, we do the same thing as in other places where we generate all the `.json` files, then we upload them as an artifact, the an `upload-test-suites` step does the actual upload.

DilumAluthge:
I prefer option 2.

staticfloat:
The only thing I'm not certain about is how easy it will be to properly "attribute" test results to a location in another step.  For instance, when using the default uploader, it [automatically uses things like the `$BUILDKITE_JOB_ID`](https://github.com/buildkite-plugins/test-collector-buildkite-plugin/blob/a3b4d39e11e7a4128d7d5321666f8bfde5d4fcbf/hooks/pre-exit#L45), which will be wrong in our case.  We need to write down all the information that buildkite needs to properly identify the step it's currently running, and then upload it from _another_ step, with that information faithfully carried over.

utilities/test_julia.sh
```diff
@@ -148,4 +148,18 @@ if [[ -z "${USE_RR-}" ]]; then
...
+# set -e; requires us using if to check the exit status
+if ${JULIA_CMD_FOR_TESTS:?} -e "Base.runtests(\"${TESTS:?}\"; ncores = ${NCORES_FOR_TESTS:?})"; then
+  exitVal=0
+else
+  exitVal=1
+fi
+
+echo "--- Upload results.json report"
+if [[ -f "results.json" ]]; then
+    buildkite-agent artifact upload "results.json"
```
IanButterworth, line 160:
```suggestion
if [[ -f "${JULIA_INSTALL_DIR}/share/julia/test/results.json" ]]; then
    buildkite-agent artifact upload "${JULIA_INSTALL_DIR}/share/julia/test/results.json"
```

vchuravy:
```
2024-02-04 04:32:02 INFO   Found 1 files that match "julia-6be4235ec9/share/julia/test/results.json"
2024-02-04 04:32:02 INFO   Uploading to default Buildkite artifact storage
2024-02-04 04:32:02 INFO   Creating (0-1)/1 artifacts
2024-02-04 04:32:02 INFO   Uploading artifact 018d7263-3d87-4994-8d4a-370246a34c04 julia-6be4235ec9/share/julia/test/results.json (3 bytes)
2024-02-04 04:32:03 INFO   Successfully uploaded artifact "julia-6be4235ec9/share/julia/test/results.json"
2024-02-04 04:32:05 INFO   Artifact uploads completed successfully
```

:partying_face:

vchuravy:
Uploading them as an artifact is the first step, we then need to do the processing and uploading https://github.com/JuliaCI/julia-buildkite/pull/150#pullrequestreview-1018875195 and https://buildkite.com/docs/test-analytics/importing-json

So we need a pipeline that can access our secrets and then downloads the results.json from the previous steps.

IanButterworth:
It's not an empty file here https://buildkite.com/julialang/julia-buildkite/builds/1426#018d7241-755b-4dfa-9552-13c46b244c91/1443-1447

vchuravy:
Let's merge this PR as is so that we can work our the kinks in the report format? I see a lot of unknown in there

IanButterworth:
SGTM

PR #153, opened 2022-06-25
Move the Apple Silicon tester from "allow fail" to regular (to prepare for promoting Apple Silicon to Tier 1)

DilumAluthge:
This is the sequel to #149.


vchuravy:
With https://buildkite.com/julialang/julia-master/builds/13291#0181ab46-dd2d-4df8-ba66-4f2de05e60d3 we should now be ready

PR #154, opened 2022-06-25
Only upload rr traces if the process fails

DilumAluthge:
I'm not sure if this PR is a good idea.

Pro: Because we don't upload rr traces if the process succeeds, we can save a bit of CI time.

Con: If something breaks in the rr upload script, we won't notice until a test fails, which potentially could cause us to fail to upload the rr trace of a rare nondeterministic bug.


staticfloat:
I don't think this is worth it; it's also useful to have "good" `rr` traces so that when something breaks, we can compare it to a recent good run's `rr` trace.

PR #155, opened 2022-06-26
Automatically retry the macOS tests for a maximum of `1` retry (but only if the exit code is `143`, and only if the build is not a pull request build)

DilumAluthge:


utilities/arches_pipeline_upload.sh
```diff
@@ -22,7 +22,16 @@ fi
...
-    # Export the environment mappings, then launch the yaml file
+    # 1. Export the environment mappings,
     eval "export ${env_map}"
+
+    # 2. Enable automatic retries if and only if this is not a pull request build.
+    if [[ "${BUILDKITE_PULL_REQUEST:?}" == "false" ]] ; then
+        export AUTOMATIC_RETRY_LIMIT=1
+    else
+        export AUTOMATIC_RETRY_LIMIT=0
+    fi
```
staticfloat, line 33: I think this logic belongs more in `arches_env.sh`

staticfloat (changes requested):

DilumAluthge:
Mac is far more stable now (as is CI in general).

PR #156, opened 2022-06-26
Increase the timeout for the macOS x86_64 tests from `90` minutes to `180` minutes

DilumAluthge:


DilumAluthge:
@staticfloat Is this the correct way to increase the timeout? And have I increased it to the value that I specified in the PR title? The timeout code has been refactored a bit, and I'm not sure where the e.g. `45` minute cleanup value is defined.

DilumAluthge:
The macOS x86_64 tests are still timing out at 90 minutes, so clearly I'm not modifying the timeout value correctly.

@staticfloat:

1. What is the correct way to increase/decrease the timeout for a test job?
2. Should we try to document this in the devdocs?

PR #158, opened 2022-06-26
Print kernel version and compiler versions

gbaraldi:


utilities/build_julia.sh
```diff
@@ -10,6 +10,10 @@ set -euo pipefail
 source .buildkite/utilities/build_envs.sh
 source .buildkite/utilities/word.sh
 
+echo "--- Get kernel/compiler versions"
+uname -a
+cc -v
+ld -v
```
DilumAluthge, line 16: Would be nice to have a blank line after this line and before the next section starts.

utilities/build_julia.sh
```diff
@@ -10,6 +10,10 @@ set -euo pipefail
 source .buildkite/utilities/build_envs.sh
 source .buildkite/utilities/word.sh
 
+echo "--- Get kernel/compiler versions"
```
DilumAluthge, line 13:
```suggestion
echo "--- Print kernel and compiler versions"
```

utilities/test_julia.sh
```diff
@@ -10,6 +10,11 @@ set -euo pipefail
 # First, get things like `SHORT_COMMIT`, `JULIA_CPU_TARGET`, `UPLOAD_TARGETS`, etc...
 source .buildkite/utilities/build_envs.sh
 
+echo "--- Get kernel/compiler versions"
```
DilumAluthge, line 13:
```suggestion
echo "--- Print kernel and compiler versions"
```

utilities/build_julia.sh
gbaraldi: You mean like print a new line? Because in the log they are in a different section already

DilumAluthge: No, just a blank line in the shell script, just for readability.

gbaraldi: Oh :), I was looking at the wrong script and was confused because it already had a newline.

DilumAluthge (approved):

DilumAluthge:
cc: @Keno @giordano

utilities/test_julia.sh
```diff
@@ -10,6 +10,11 @@ set -euo pipefail
 # First, get things like `SHORT_COMMIT`, `JULIA_CPU_TARGET`, `UPLOAD_TARGETS`, etc...
 source .buildkite/utilities/build_envs.sh
 
+echo "--- Print kernel and compiler versions"
+uname -a
+cc -v
+ld -v
```
giordano, line 16: Are these the actual tools used for compilation?

giordano: If we have to trust https://github.com/JuliaCI/rootfs-images/blob/57890bdbfc7273c0d88ef527fbfc43766e99135f/linux/package_linux.jl#L56 we should see GCC 9.1, but https://buildkite.com/julialang/julia-buildkite/builds/687#0181a1ff-4eb2-4919-9599-010db739d635/404-412 shows
```
gcc version 8.3.0 (Debian 8.3.0-6)
```

DilumAluthge: @staticfloat Should https://github.com/JuliaCI/rootfs-images/blob/main/linux/package_linux.jl be creating a symlink for `cc` -> ... (in addition to the `gcc` -> ... symlink that we already create)?

giordano: https://buildkite.com/julialang/julia-buildkite/builds/688#0181a20b-c97c-44ae-b573-5ea80b4e3fe6/413-421
```
gcc version 9.3.0 (Alpine 9.3.0)
```
musl is not using the compiler from the RootFS?

DilumAluthge: Here's the RootFS for musl: https://github.com/JuliaCI/rootfs-images/blob/main/linux/package_musl.jl

gbaraldi: cc is different from gcc. Not sure what is actually called in the makefiles. In macos it doesn't make a difference since they all are linked to clang in the end

DilumAluthge: `package_linux.jl` is only for glibc Linux.

staticfloat: PR to create `cc` and `c++` symlinks: https://github.com/JuliaCI/rootfs-images/pull/209

staticfloat: We don't use compilers in the `tester_` things.  There is no guarantee that you'll have `cc` available to you.  We can dump kernel version here, but the compiler versions don't mean anything.

PR #164, opened 2022-06-29
Move all scheduled jobs to a dedicated "scheduled" pipeline

### LINKED ISSUE #157: TODO: move the scheduled jobs to a dedicated "scheduled" pipeline
Originally, the once-daily scheduled jobs went into their own pipeline. In other words, we had:
1. `julia-master`: the main pipeline. Contained only the non-scheduled jobs.
2. `julia-master->scheduled`: the scheduled pipeline. Contained only the scheduled jobs.

Recently (in #102), I moved all the scheduled jobs into the main pipeline and deleted the scheduled pipeline. So, currently we have:
1. `julia-master`: the main pipeline. Contain all jobs (both non-scheduled and scheduled).

At the time, I thought this was a good idea. It turns out that it is a bad idea. The non-scheduled builds and scheduled builds are trampling all over each other's GitHub commit statuses. If at 1AM a regular (non-scheduled build) is triggered on a push to master, and all three of our groups/GitHub commit statuses (`Build`, `Check`, `Test`) are green on that commit, and then at 5AM a scheduled build is triggered on that exact same commit, now the scheduled build will erase those three green commit statuses.

Also, I find it quite confusing that when I am scrolling through https://buildkite.com/julialang/julia-master/builds?branch=master I see a mix of non-scheduled and scheduled builds, which makes it difficult when I am e.g. trying to figure out when a breakage first started occuring on master.

So it turns out that it will be better to have two separate pipelines, the main `julia-master` pipeline and the scheduled `julia-master->scheduled` pipeline. As I wrote in #102, yes it is annoying to maintain two separate pipelines, but as I have now learned, it is even more annoying (and confusing, inconvenient, etc.) to have them combined into a single pipeline. So we will go with the lesser of two evils.

DilumAluthge:
Fixes #157. See #157 for the motivation.

## List of Buildkite pipelines

- **Main pipeline**:
    - JuliaLang/julia:
        1. https://buildkite.com/julialang/julia-master
        2. https://buildkite.com/julialang/julia-release-1-dot-6
        3. https://buildkite.com/julialang/julia-release-1-dot-7
        4. https://buildkite.com/julialang/julia-release-1-dot-8
        5. [...]
    - JuliaCI/julia-buildkite:
        1. https://buildkite.com/julialang/julia-buildkite
- **Scheduled pipeline**:
    - JuliaLang/julia:
        1. https://buildkite.com/julialang/julia-master-scheduled
    - JuliaCI/julia-buildkite:
        1. https://buildkite.com/julialang/julia-buildkite-scheduled


DilumAluthge:
@staticfloat This is good to merge as soon as the `buildkite/julia-buildkite-scheduled` status turns green on this PR.

Then, rebase all existing PRs, and the `buildkite/julia-buildkite-scheduled` status will turn green on those PRs as well.

pipelines/main/platforms/upload_linux.arches
```diff
@@ -1,9 +1,9 @@
-# OS       TRIPLET                  TIMEOUT
```
giordano, line 1: This broke the upload jobs?

DilumAluthge: Yes.

DilumAluthge: https://github.com/JuliaCI/julia-buildkite/pull/170

PR #165, opened 2022-06-29
Set JULIA_TEST_TIMEOUT_SIGNUM=11

staticfloat:
Let's try to get some coredumps from our timeouts


staticfloat:
> Any particular reason to use SIGSEGV instead of SIGQUIT?

Nope, that's a great idea.

staticfloat:
@DilumAluthge are you messing with the scheduled builder?  I got a PR status from it here...

DilumAluthge:
> @DilumAluthge are you messing with the scheduled builder? I got a PR status from it here...

Yeah. You can safely ignore this failure. It will be fixed once I merged #164.

staticfloat:
We're waiting on this until I understand signal handling better, and am able to reliably trigger core dumps on all platforms.

DilumAluthge:
Okay, so we made some progress by identifying that macOS was resetting our coredump config on reboot.

I think we also need to change by to using SIGSEGV, because as Jameson said on Slack, Julia intercepts SIGQUIT, and thus SIGQUIT won't trigger core dumps.

DilumAluthge:
@staticfloat Can we try to get this in?

DilumAluthge:
With https://github.com/JuliaLang/julia/pull/47234, I think that if the SIGSEGV is sent by us via `kill`, then Julia should hard fail (instead of catching the signal and turning it into a Julia exception), and we should get our desired core dumps.

PR #170, opened 2022-07-10
Fix the `upload_linux.arches` file

DilumAluthge:


DilumAluthge:
Looks like "Launch upload jobs" is fixed.

PR #179, opened 2022-07-20
Limit `tester_linux` builds to run on `cpuset_limited=true` agents

staticfloat:
This will allow us to create multiple non-cpuset-limited workers, which
will allow us to do things like builds with oversubscribed CPUs, but
tests with pinned, dedicated CPUs.

### DIFF
--- a/pipelines/main/platforms/test_linux.32.yml
+++ b/pipelines/main/platforms/test_linux.32.yml
@@ -39,6 +39,7 @@ steps:
           sandbox_capable: "true"
           os: "linux"
           arch: "${ARCH?}"
+          cpuset_limited: "true"
         env:
           JULIA_SHELL: "/bin/bash"
           TRIPLET: "${TRIPLET?}"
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Before we can merge this PR, we first need to add some agents that have both `arch=powerpc64le` and `cpuset_limited=true`.

DilumAluthge:
> Before we can merge this PR, we first need to add some agents that have both `arch=powerpc64le` and `cpuset_limited=true`.

@staticfloat Did you have a chance to add an agent that satisfies these criteria?

staticfloat:
[Yep](https://buildkite.com/organizations/julialang/agents?q=power+cpuset)

DilumAluthge:
@staticfloat Is this good to merge?

staticfloat:
Yep!  This won't have much of an impact at first, but we'll add some non-cpuset-limited builders in the future, and those will be responsible for running our builds.

PR #181, opened 2022-07-22
At the end of each test job, upload all files in the `JULIA_TEST_VERBOSE_LOGS_DIR` directory as Buildkite artifacts

DilumAluthge:
When testing Julia:

1. Do not set the `JULIA_PKG_TEST_QUIET` environment variable.
2. Create a temporary directory and set the `JULIA_TEST_VERBOSE_LOGS_DIR` environment variable to that directory.
3. At the end of each test job, whether or not the tests pass, upload all files in the `JULIA_TEST_VERBOSE_LOGS_DIR` directory as Buildkite artifacts.

### DIFF
--- a/utilities/test_julia.sh
+++ b/utilities/test_julia.sh
@@ -40,9 +40,6 @@ export OPENBLAS_NUM_THREADS="${JULIA_CPU_THREADS}"
 unset JULIA_DEPOT_PATH
 unset JULIA_PKG_SERVER
 
-# Show the output from the Pkg tests
-export JULIA_PKG_TEST_QUIET="false"
-
 # Make sure that temp files and temp directories are created in a location that is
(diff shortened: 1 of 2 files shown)


DilumAluthge:
@staticfloat Is `utilities/timeout.jl` the correct place to put this logic

gbaraldi:
We could do it in test_julia.sh instead. But this is fine IMO.

DilumAluthge:
@staticfloat @gbaraldi Looks like the Pkg logs are available as Buildkite artifacts. E.g. go to https://buildkite.com/julialang/julia-buildkite/builds/840#0182481a-6ee0-4737-b964-a1a3175d77ea and click on the Artifacts tab, and then you can click on the `Pkg.log` file.

staticfloat (approved):

PR #195, opened 2022-08-22
Do not test FileWatching on macos

gbaraldi:
Macos does some odd stuff with event handling, I tried disabling some of the main culprit tests on the julia repo, but it seems they are all unreliable so I'd rather disable them all while we find some solution to it.

Maybe test them during the nightly run only just to keep coverage?


staticfloat:
Closing as no longer needed, per Gabriel's comments to me on slack.

PR #196, opened 2022-08-23
When running the tests, set the `JULIA_TEST_IS_BASE_CI` environment variable to `true`

DilumAluthge:


DilumAluthge:
@gbaraldi In the FileWatching tests, you can skip the FileWatching tests if `(get(ENV, "JULIA_TEST_IS_BASE_CI", "" == "true") && Sys.isapple()`.

PR #199, opened 2022-09-11
Build the PDF docs on Buildkite

DilumAluthge:
Ref https://github.com/JuliaCI/julia-buildkite/issues/198

### DIFF
--- a/pipelines/main/misc/pdf_docs/build_pdf_docs.yml
+++ b/pipelines/main/misc/pdf_docs/build_pdf_docs.yml
@@ -22,15 +22,11 @@ steps:
                 # Include `/cache/repos` so that our `git` version introspection works.
                 - "/cache/repos:/cache/repos"
         commands: |
-          # echo "--- Download pre-built julia, extract into usr/"
-          # buildkite-agent artifact download --step "build_x86_64-linux-gnu" 'julia-*-linux-x86_64.tar.gz' .
-          # mkdir -p ./usr
-          # tar -C ./usr --strip-components=1 -zxf julia-*-linux-x86_64.tar.gz
-          # ln -s ./usr/bin/julia ./julia
-
-          echo "--- Build Julia"
-          make -j$${JULIA_CPU_THREADS:?}
-          make install
+          echo "--- Download pre-built julia, extract into usr/"
+          buildkite-agent artifact download --step "build_x86_64-linux-gnu" 'julia-*-linux-x86_64.tar.gz' .
+          mkdir -p ./usr
+          tar -C ./usr --strip-components=1 -zxf julia-*-linux-x86_64.tar.gz
+          ln -s ./usr/bin/julia ./julia
 
           echo "--- Print Julia version info"
           export JULIA_NUM_THREADS=1
--- a/pipelines/main/misc/pdf_docs/build_pdf_docs.yml
+++ b/pipelines/main/misc/pdf_docs/build_pdf_docs.yml
@@ -1,0 +1,55 @@
+steps:
+  - group: "Check"
+    steps:
+      - label: "pdf_docs_build"
+        key: pdf_docs_build
+        # depends_on:
+        #   - "build_x86_64-linux-gnu"
+        plugins:
+          - JuliaCI/external-buildkite#v1:
+              version: "./.buildkite-external-version"
+              repo_url: "https://github.com/JuliaCI/julia-buildkite"
+          - JuliaCI/julia#v1:
+              # Drop default "registries" directory, so it is not persisted from execution to execution
+              persist_depot_dirs: packages,artifacts,compiled
+              version: '1'
+          - staticfloat/sandbox#v1:
+              rootfs_url: https://github.com/JuliaCI/rootfs-images/releases/download/v5.38/latex.x86_64.tar.gz
+              rootfs_treehash: "be2d12d899ee334c470a79f6d3f3faa4e63c7c3c"
+              uid: 1000
+              gid: 1000
+              workspaces:
+                # Include `/cache/repos` so that our `git` version introspection works.
+                - "/cache/repos:/cache/repos"
+        commands: |
+          # echo "--- Download pre-built julia, extract into usr/"
+          # buildkite-agent artifact download --step "build_x86_64-linux-gnu" 'julia-*-linux-x86_64.tar.gz' .
+          # mkdir -p ./usr
+          # tar -C ./usr --strip-components=1 -zxf julia-*-linux-x86_64.tar.gz
+          # ln -s ./usr/bin/julia ./julia
+
+          echo "--- Build Julia"
+          make -j$${JULIA_CPU_THREADS:?}
+          make install
+
+          echo "--- Print Julia version info"
+          export JULIA_NUM_THREADS=1
+          ./julia -e 'using InteractiveUtils; InteractiveUtils.versioninfo()'
+
+          echo "--- Set some environment variables and create some directories"
+          export DOCUMENTER_LATEX_DEBUG="$$(pwd)/latex-debug-logs"
+          mkdir -p $${DOCUMENTER_LATEX_DEBUG:?}
+          echo "DOCUMENTER_LATEX_DEBUG is $${DOCUMENTER_LATEX_DEBUG:?}"
+
+          echo "--- Build the PDF"
+          ./julia .buildkite/pipelines/main/misc/pdf_docs/make.jl
+
+          echo "--- Upload the PDF files as Buildkite artifacts"
+          find . -type f -name "*.pdf"
+          find . -type f -name "*.pdf" -exec buildkite-agent artifact upload {} \;
+        timeout_in_minutes: 120
+        agents:
+          queue: "julia"
+          sandbox_capable: "true"
+          os: "linux"
+          arch: "x86_64"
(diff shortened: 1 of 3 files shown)


DilumAluthge:
@mortenpi After I run `make.jl`, where does it save the PDF files?

mortenpi:
> After I run `make.jl`, where does it save the PDF files?

I think the `make.jl` copies it into `JULIA_DOCS_TMP`.

However, I am not sure that it makes sense to borrow the `make.jl` script from the other repo at all. Couldn't we just call

```
make -C $(JULIA_SOURCE)/doc pdf texplatform=docker JULIA_EXECUTABLE=$(julia_exec)
```

in the YAML and then pass the PDF on to the `deploy_docs` pipeline?

The latter would also have to be modified a little of course. The PDFs get deployed onto the `assets` branch of the repo, and we also overwrite the history on every push.

Also, cc @fredrikekre since he's more familiar with the PDF deployment.

DilumAluthge:
@staticfloat It's failing to load the rootfs. Maybe it's running out of disk space?

pipelines/main/misc/pdf_docs/build.yml
```diff
@@ -0,0 +1,63 @@
...
+          - JuliaCI/external-buildkite#v1:
+              version: "./.buildkite-external-version"
+              repo_url: "https://github.com/JuliaCI/julia-buildkite"
+          - JuliaCI/julia#v1:
+              # Drop default "registries" directory, so it is not persisted from execution to execution
+              persist_depot_dirs: packages,artifacts,compiled
+              version: '1'
+          - staticfloat/sandbox#v1:
+              rootfs_url: https://github.com/JuliaCI/rootfs-images/releases/download/v5.32/python_latex_kitchen_sink.x86_64.tar.gz
+              rootfs_treehash: "0000000000000000000000000000000000000000"
```
staticfloat, line 18: I assume you need to provide an actual treehash here?  :P

DilumAluthge:
What are the possible values for `texplatform`? Docker isn't available here, but we do have LaTeX already installed in this rootfs image.

mortenpi:
I guess as you figured out, the `texplatform=..` part can be omitted, in which case it uses the "native" backend (i.e. local installation). Do we have the full distro on rootfs? And it also requires some custom fonts to be installed.

Here is the Dockerfile that sets up the Docker image: https://github.com/JuliaDocs/DocumenterLaTeX.jl/blob/master/docker/Dockerfile

DilumAluthge:
We should have everything we need in the existing rootfs image, but we can make a new one if need be.

Right now the build is failing, but I'm not sure what the cause is.

mortenpi:
What's the working directory [here](https://buildkite.com/julialang/julia-buildkite/builds/890#01832eb2-f263-4908-81ac-4b5b65e8b3f6/476-485)? I see on 477 that `make: Entering directory '/cache/build/default-amdci5-1/julialang/julia-buildkite/doc` which is different from [573](https://buildkite.com/julialang/julia-buildkite/builds/890#01832eb2-f263-4908-81ac-4b5b65e8b3f6/476-573) where `SystemError: opening file "/cache/build/default-amdci5-1/julialang/julia-buildkite/pdf/build/doc/UnicodeData.txt"`.

DilumAluthge:
Oh maybe it's because I set the `BUILDROOT` environment variable. Let me comment that out.

pipelines/main/misc/build_pdf_docs.yml
```diff
@@ -0,0 +1,62 @@
...
+          echo "DOCUMENTER_LATEX_DEBUG is $${DOCUMENTER_LATEX_DEBUG:?}"
+          echo "JULIA_EXECUTABLE is $${JULIA_EXECUTABLE:?}"
+          echo "JULIA_SOURCE is $${JULIA_SOURCE:?}"
+
+          echo "--- Make PDF"
+          make -C doc pdf JULIA_EXECUTABLE=$${JULIA_EXECUTABLE:?}
+
+          echo "--- Upload the latex debug logs"
+          tar czvf latex-debug-logs.tar.gz $${DOCUMENTER_LATEX_DEBUG}
+          buildkite-agent artifact upload latex-debug-logs.tar.gz
```
mortenpi, line 52: I guess this part didn't run when `make` failed?

mortenpi:
Alright, it looks like it complains about [Pygments](https://pygments.org/docs/cmdline/) being missing:

```
! Package minted Error: You must have `pygmentize' installed to use this package.
```

`sudo apt install python-pygments` ought to be enough for this one. But after that we'll probably still be missing some fonts.

DilumAluthge:
We'll likely need to make a separate rootfs image just for this job.

https://github.com/JuliaCI/rootfs-images/pull/224

mortenpi:
Hmm.. is `make` missing from the new rootfs image? And Documenter also needs `git`, which I think might also be missing?

DilumAluthge:
https://github.com/JuliaCI/rootfs-images/pull/225

DilumAluthge:
Okay, it looks like the PDF was built successfully.

@mortenpi Would you mind looking at the PDF (which was uploaded as a Buildkite artifact) and making sure it looks correct?

DilumAluthge:
@staticfloat In this PR, we build Julia in the `build_pdf_docs` job. I had trouble getting the "use a pre-built Julia" to work.

In a follow-up PR, we can switch the `build_pdf_docs` job to using a pre-built Julia, but for this PR, it might be easier to leave it as-is so that we can get this PR merged relatively quickly.

pipelines/main/misc/pdf_docs/build_pdf_docs.yml
```diff
@@ -0,0 +1,55 @@
...
+                # Include `/cache/repos` so that our `git` version introspection works.
+                - "/cache/repos:/cache/repos"
+        commands: |
+          # echo "--- Download pre-built julia, extract into usr/"
+          # buildkite-agent artifact download --step "build_x86_64-linux-gnu" 'julia-*-linux-x86_64.tar.gz' .
+          # mkdir -p ./usr
+          # tar -C ./usr --strip-components=1 -zxf julia-*-linux-x86_64.tar.gz
+          # ln -s ./usr/bin/julia ./julia
+
+          echo "--- Build Julia"
```
giordano, line 31: Why does this need to build julia? Can't it use the existing binary from a build on the same trigger?

DilumAluthge: Yes, it should. I was just running into weird errors when I tried that initially. (`make` was complaining that it couldn't find a Fortran compiler.)

So I switched to building Julia just to get this PR working quickly, so that we can merge it and start using Buildkite to deploy the PDF manual.

staticfloat:
It should be identical to the `doctest` builder.  Also, why aren't we just adding this on to the end of the `doctest` builder?  Because we need `pdflatex`?  Why not try switching the `doctest` builder to the new `latex` rootfs you have here and building the HTML and PDF docs at once?

DilumAluthge:
> Also, why aren't we just adding this on to the end of the `doctest` builder?

This was intentional. It seems to me that the PDF build currently breaks a lot. And I didn't want a broken PDF build to prevent the regular (i.e. HTML) docs from being deployed. So I figured we'd keep them separate, at least initially, so that we can monitor the stability of the PDF build for a while.

mortenpi:
The PDF looks fine to me! I think we just now need a variant of the  `deploy_docs.yml` job to actually push the PDF to the `assets` branch?

> This was intentional. It seems to me that the PDF build currently breaks a lot. And I didn't want a broken PDF build to prevent the regular (i.e. HTML) docs from being deployed.

I think this is a fair concern. I suppose they could be in the same builder if it is somehow possible to split things into independent-ish steps on Buildkite?

pipelines/main/misc/pdf_docs/build_pdf_docs.yml
```diff
@@ -0,0 +1,51 @@
...
+          export DOCUMENTER_LATEX_DEBUG="$$(pwd)/latex-debug-logs"
+          mkdir -p $${DOCUMENTER_LATEX_DEBUG:?}
+          echo "DOCUMENTER_LATEX_DEBUG is $${DOCUMENTER_LATEX_DEBUG:?}"
+
+          echo "--- Build the PDF"
+          ./julia .buildkite/pipelines/main/misc/pdf_docs/make.jl
+
+          echo "--- Upload the PDF files as Buildkite artifacts"
+          find . -type f -name "*.pdf"
+          find . -type f -name "*.pdf" -exec buildkite-agent artifact upload {} \;
```
staticfloat, line 45: Are we uploading too many `.pdf` files?

image

pipelines/main/misc/pdf_docs/build_pdf_docs.yml
DilumAluthge: We are now using a pre-built Julia.

pipelines/main/misc/pdf_docs/build_pdf_docs.yml
DilumAluthge, line 45:
```suggestion
          cd doc/_build/pdf/en && buildkite-agent artifact upload TheJuliaLanguage.pdf
```

DilumAluthge: Should be resolved by https://github.com/JuliaCI/julia-buildkite/pull/199/commits/efeb5dff3cebd304ea80c2ffcfc806e84ca5df26

DilumAluthge:
The core functionality is here. We use a pre-built Julia, build the PDF, and upload it as a Buildkite artifact.

I'm going to merge this now so that we can start running it on Base Julia (and thus catch any regressions). The signed pipeline to upload the PDF to the correct GitHub repo can be done in a follow-up PR.

Issue #25, opened 2022-01-28
Meta-issue: Windows


DilumAluthge:
Current status:

> Windows - These required a lot of work because our windows buildbot story has always been a :dumpster-fire: .  Previously, we had someone build windows VMs, used cygwin to install a bunch of compiler versions, and then whatever was installed was just what we got.  Updating the system meant potentially upgrading the compiler versions and breaking everything.  It was awful.  Also, the windows VMs often get wedged and must be restarted.  I have recently managed to streamline Windows VM creation via Packer, and figured out the proper KVM recipes to be able to run docker-on-windows-in-KVM on our amdci machines.  This means that we can build the system toolchains as docker images that get deployed onto the machines, providing reproducibility, and we can destroy the docker container after each build, which should provide a lot more reliability.  We will definitely need help creating the windows docker images and spearheading buildkite windows development, as I believe it’s mostly in powershell.

Next steps:

> We need to build windows docker images that contain the mingw32 build tools and windows platform SDK headers and whatnot necessary to build Julia.  We need to also build a set of powershell instructions to build and package Julia inside of that docker container, and then integrate the whole thing into buildkite.  This will involve some powershell scripting and some buildkite plugin development, I bet.

inkydragon:
So, we want something like https://github.com/cirruslabs/docker-images-windows (windows + VisualStudio2022),
but we want w64-mingw32 not VS2022, right?

inkydragon:
Some progress: https://github.com/inkydragon/msys2-mingw-w64
I've built a windows docker image (20H2 + MSYS2 + x86_64-w64-mingw32).
This is a POC, there are still many things todo.

staticfloat:
This is great!

staticfloat:
Okay, first some feedback:

* It appears that the images, as they are, can't be run on my windows buildbots.  The buildbots are running Windows Server 2022, and I think the issue is that by default, your images are built with the `20H2` image, which is somehow incompatible (I get the error message `The container operating system does not match the host operating system.`).  I can fix this by building the image locally, and manually setting `WIN_VERSION=ltsc2022`.

* It's not 100% necessary, but it would be nice to specifically install GCC 9, rather than just the latest available version of GCC (10 right now, it appears).  This would reduce variation across OSes in our build fleet, and also makes it less likely for us to rebuild the image and things suddenly break.

* When trying to build Julia, bootstrap fails with the following:
[...]

PR #200, opened 2022-09-14
Add windows build

staticfloat:
After integrating the great work done by @inkydragon in https://github.com/JuliaCI/julia-buildkite/issues/25 into https://github.com/JuliaCI/rootfs-images/pull/227 and running https://github.com/JuliaCI/sandboxed-buildkite-agent/pull/22 we are ready to try this out.


DilumAluthge:
@staticfloat Instead of pulling the Docker image by the `v*.*` tag, what do you think about pulling by the sha256 digest (which can be found at the bottom of the CI logs)?

staticfloat:
I think that makes it much more difficult to, at a glance, figure out what version we're actually using.  Why wouldn't you want to use the tag name?

DilumAluthge:
> I think that makes it much more difficult to, at a glance, figure out what version we're actually using. Why wouldn't you want to use the tag name?

Only that digests are immutable and tags are mutable. So it protects us from breaking CI by accidentally pushing to the same tag.

That would be an unlikely event, though, so preventing it might not be worth the loss of readability.

staticfloat:
Yeah; even if we did push to the same tag, it would just download the new image, which IMO is not that big of a deal.  If we didn't control the images that are being pushed up to docker hub, I would be more interested in this, but since we control it, I figure using the human-readable name is best.

staticfloat:
Uhghhhhh the win32 compiler uses the wrong exception model.  It's linking against `libgcc_s_dw2-1.dll`.

staticfloat:
With a new compiler, and some massaging, we've got builds that are running tests!  Some things to look into:

- [ ] `SystemError: GetClipboardData: Invalid clipboard format`: [Example](https://buildkite.com/julialang/julia-buildkite/builds/952#01836695-a144-48ea-a452-23665617dbb7/358-817).  Possbily some consequence of [how we're starting the buildkite agent](https://github.com/JuliaCI/sandboxed-buildkite-agent/blob/8567e1a2d18ae0b0a371f20d1a8f878d9e242922/windows-kvm/buildkite-worker/setup_scripts/0-02-install-buildkite-agent.ps1#L12-L16) in a headless fashion?
- [ ] `A "spawn and wait lots of tasks" test failed`: [Example](https://buildkite.com/julialang/julia-buildkite/builds/952#01836695-a144-48ea-a452-23665617dbb7/358-967).  Sounds like a segfault, but unsure why.

DilumAluthge:
> * We should disable things that rely on backtraces on win32, such as the `Profile` test, since that is extremely slow there.

cc: @ianbutterworth, who may have some hints/tips for us here.

staticfloat:
I think this is good to go!  @DilumAluthge want to take a quick look over it?

DilumAluthge (approved):
So the build runs inside Docker, but the tests do not? Any particular reason for this?

Otherwise LGTM. Agree with leaving the tests as "allow fail" initially.

staticfloat:
> So the build runs inside Docker, but the tests do not? Any particular reason for this?

Because we don't need (and indeed, don't want) any compilers around for the tests.  There's enough basic utilities available in the base image that we don't need a docker image for the tests.

DilumAluthge (approved):

PR #204, opened 2022-09-23
Don't double-timeout `rr` tests

staticfloat:
`rr_capture.jl` already contains within it timeout logic; so let's not kill it right when it would otherwise collect a backtrace.


DilumAluthge:
@staticfloat Fixes https://github.com/JuliaLang/julia/issues/46859?

DilumAluthge (approved):

PR #206, opened 2022-09-26
Add a once-daily scheduled job to build Julia with `USE_GPL_LIBS=0`

DilumAluthge:
Currently only covers a few architectures, specificallly:

1. x86_64-linux-gnu
2. x86_64-apple-darwin
3. aarch64-apple-darwin


pipelines/scheduled/launch_unsigned_jobs.yml
```diff
@@ -35,3 +35,26 @@ steps:
               .buildkite/pipelines/main/platforms/test_linux.yml
         agents:
           queue: "julia"
+  - group: "LGPL"
```
giordano, line 38: Out of curiosity, what LGPL stands for here?

DilumAluthge: It's to reflect the fact that if you do `USE_GPL_LIBS=0`, it builds Julia without the GPL dependencies, but it still includes the LGPL dependencies (GMP, MPFR, etc.). So the resulting binaries are not subject to the GPL, but they are still subject to the LGPL license.

(I am not a lawyer, etc.)

giordano: From what I can tell from https://github.com/JuliaLang/julia/blob/d97f2483eaca36f9d65ddb3c444ab52d43895fc8/THIRDPARTY.md there are also files/libraries with a bunch of other licenses, like MIT, Apache, boost, etc.

DilumAluthge: Maybe just "NO_GPL" instead?

giordano: Yes, I think that'd convey the message more clearly.

PR #210, opened 2022-09-28
Codesign and notarize the "no GPL" macOS and Linux binaries and upload them to S3

DilumAluthge:


DilumAluthge:
There is good news and bad news.

The good news: I was able to download the `.tar.gz` file from https://s3.amazonaws.com/julialang-ephemeral/julia-buildkite-uploads/bin/macosnogpl/x86_64/1.9/julia-9fd408723a-macosnogpl-x86_64.tar.gz

The bad news: when I try to run `./bin/julia`, I get the following popup. I think that means it's not codesigned or notarized propertly.

Screen Shot 2022-09-30 at 7 36 07 PM

DilumAluthge:
@staticfloat Any idea why the notarization didn't seem to work?

DilumAluthge:
I think this is working correctly:

1. I was able to download the `.tar.gz` file from https://s3.amazonaws.com/julialang-ephemeral/julia-buildkite-uploads/bin/macosnogpl/x86_64/1.9/julia-6424f790f4-macosnogpl-x86_64.tar.gz
2. When I extract the tarball and run `./julia-6424f790f4/bin/julia`, Julia starts up without any issues. I don't get any security pop-ups.

DilumAluthge:
I'm going to go ahead and merge this, but @staticfloat would you mind reviewing this when you get a chance, just to make sure that I did everything correctly?

PR #214, opened 2022-09-30
Rework test selection

staticfloat:
This builds two bash arrays, `TESTS_TO_RUN` and `TESTS_TO_SKIP`, which can be added to more flexibly than just building a monolithic `TESTS` string variable.  We then adjust these variables within our various OS or ARCH-specific conditionals, finally building `TESTS` at the end.

This also adjusts win32 to skip the Profile tests, as they are extremely slow because our stack walker is slow on win32, and we're not going to fix that any time soon.

### DIFF
--- a/utilities/test_julia.sh
+++ b/utilities/test_julia.sh
@@ -52,7 +52,11 @@ if [[ "${OS}" == "linux" ]]; then
     mkdir -p ${TMPDIR}
 fi
 
-# If we're running inside of `rr`, limit the number of threads
+# By default, we'll run all tests and skip nothing
+TESTS_TO_RUN=( "all" )
+TESTS_TO_SKIP=()
+
+# If we're running inside of `rr`, limit the number of threads and split our tests
 if [[ "${USE_RR-}" == "rr" ]] || [[ "${USE_RR-}" == "rr-net" ]]; then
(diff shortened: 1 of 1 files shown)


DilumAluthge (changes requested):
Hmmmm. I personally find this harder to read, i.e. it's hard for me to figure out at a quick glance what tests will be run.

staticfloat:
I think it's much cleaner to just keep track of which tests to run and which to skip, otherwise you have to do things like [this](https://github.com/JuliaCI/julia-buildkite/blob/main/utilities/test_julia.sh#L84) where the conditional is supposed to be related to `Downloads`, but we need to stick `Pkg` in there because we're overwriting what was [previously set on line 78](https://github.com/JuliaCI/julia-buildkite/blob/main/utilities/test_julia.sh#L78).  I feel the "overwriting" behavior is much more confusion.

DilumAluthge:
Yeah, I agree that the old system is also hard to follow.

Let me mull this over a bit more and see if I can come up with something different?

DilumAluthge (approved):
I've re-reviewed this. I think this is a improvement over the status quo. I do think there might be some tweaks I want to make in the future, but let's get this merged, because it's better than what we have.

PR #218, opened 2022-10-05
[bughunt] Add build_julia and test_julia commands

staticfloat:
This teaches the `bughunt` environment how to use the scripts in
`.buildkite/utilities` to build and test Julia, so that we can easily
and 100% reproducibly recreate the build and test invocations that we
have on CI. It even checks out the same `julia-buildkite` commit as was
used in the build, in case we changed something!

In order to do this, we are creating a fake `buildkite-agent` script to
mock out some of the interactions we have with `buildkite-agent` in the
test scripts.  I am also feeding in a whole lot more of the environment
from buildkite, and substituting in a few other things, such as the
number of cores in the current machine.

When you use the `test_julia` script, it will use a from-source-built
Julia if you made one with `build_julia`, otherwise it will prefer to
use the prebuilt julia downloaded from the test step, if one exists.


debugging/bughunt/bughunt.jl
```diff
@@ -31,5 +33,5 @@ collect_resources(build_info, prefix)
 # Construct SandboxConfig off of our build info
 config = SandboxConfig(build_info, prefix)
 with_executor() do exe
-    run(exe, config, `/bin/bash -l`)
+    run(exe, config, ignorestatus(`/bin/bash -l`))
```
DilumAluthge, line 36: Could you add a comment as to why we `ignorestatus` here?

DilumAluthge (approved):

DilumAluthge:
Also CI needs to be fixed.

PR #221, opened 2022-10-07
Codesign and upload the Windows "no GPL" binaries to S3

DilumAluthge:


DilumAluthge:
@staticfloat When you have a chance, can you double-check that this PR (and also https://github.com/JuliaCI/julia-buildkite/pull/210) were both done correctly?

staticfloat:
LGTM!  I wish we didn't have to duplicate the `.yml` files so thoroughly though.

DilumAluthge:
Yeah, it's quite annoying. If we needed this as a permanent solution, I'd want to refactor the setup in some way. But once SuiteSparse and SparseArrays are removed from the standard library, the default Julia builds will be non-GPL, and thus we'll be able to delete all of these YAML files.

PR #224, opened 2022-10-29
Show build stats

IanButterworth:
I believe this used to be run.

Should help monitor sysimage size etc.


giordano:
Maybe at https://github.com/JuliaLang/julia/blob/37e0579af04919e1e6264bbfc33ed8f4c537005a/Makefile#L608-L611 the speedtest should be launched only if `time` is actually available?

DilumAluthge:
> Maybe at https://github.com/JuliaLang/julia/blob/37e0579af04919e1e6264bbfc33ed8f4c537005a/Makefile#L608-L611 the speedtest should be launched only if `time` is actually available?

It'll be easier to just add `time` to the `package_linux` and `package_musl` images: https://github.com/JuliaCI/rootfs-images/pull/240

giordano:
That doesn't change the fact that running a command which doesn't exist isn't going to work.

DilumAluthge:
Sure, but isn't `time` basically a build-time dependency for the `make build-stats` target? So it seems fine to error if the dependency is unavailable.

IanButterworth:
(feel free to push to this)

DilumAluthge:
Okay, so https://github.com/JuliaCI/rootfs-images/pull/240 will add `time` to `package_linux` and `package_musl`. From the CI logs, it looks like the macOS CI machines already have `time` installed.

So we only need to figure out Windows. Does anyone know what I need to install for Windows?

If it helps, here's how we build the `package-windows-i686` and `package-windows-x86_64` Docker images: https://github.com/JuliaCI/rootfs-images/tree/main/windows

inkydragon:
MSYS2 already has a built-in time.
```
woclass@wos-PC MINGW64 /v/julia
$ which time
/usr/bin/time
```

The main problem is that `llvm-size$(EXE)` needs the dll in the `$(build_bindir)` directory:
```
$ PATH=/v/julia/julia-mingw64/usr/bin:$PATH /v/julia/julia-mingw64/usr/tools/llvm-size.exe
V:\julia\julia-mingw64\usr\tools\llvm-size.exe: error: 'a.out': No such file or directory

$ PATH= /v/julia/julia-mingw64/usr/tools/llvm-size.exe
V:/julia/julia-mingw64/usr/tools/llvm-size.exe: error while loading shared libraries: libLLVM-14jl.dll: cannot open shared object file: No such file or directory
```

```diff
diff --git a/Makefile b/Makefile
index 84f08f4776..a5f77a7ea8 100644
--- a/Makefile
+++ b/Makefile
@@ -646,7 +646,7 @@ win-extras:
 ifeq ($(USE_SYSTEM_LLVM), 1)
 LLVM_SIZE := llvm-size$(EXE)
 else
-LLVM_SIZE := $(build_depsbindir)/llvm-size$(EXE)
+LLVM_SIZE := PATH=$(build_bindir):$$PATH; $(build_depsbindir)/llvm-size$(EXE)
 endif
 build-stats:
 ifeq ($(USE_BINARYBUILDER_LLVM),1)
```

Zentrik:
This should probably use `${MAKE}` instead of `make`.

Zentrik:
I think you need to rebase this.

Zentrik:
I think we want `${MAKE} "${MFLAGS[@]}" build-stats` for it to work for the assert builds.

utilities/build_julia.sh
```diff
@@ -66,6 +66,9 @@ echo "--- Quick consistency checks"
 ${JULIA_EXE} -e "import Test; Test.@test Sys.ARCH == :${ARCH:?}"
 ${JULIA_EXE} -e "import Test; Test.@test Sys.WORD_SIZE == ${EXPECTED_WORD_SIZE:?}"
 
+echo "--- Show build stats"
+${MAKE} build-stats
```
Zentrik, line 70:
```suggestion
${MAKE} "${MFLAGS[@]}" build-stats
```

Zentrik: bump

IanButterworth: Personally I don't understand this change but perhaps you could ping someone that might? Or explain?

Zentrik: It adds the various make options that are already used when building julia, defined here https://github.com/IanButterworth/julia-buildkite/blob/9fcbe0713621dfec83f0c621c1c5fa3fca5ea82f/utilities/build_julia.sh#L23-L45.

This fixes the current error with assert builds as it will now build with the `MAKE_FLAGS` defined in the .arches file, i.e. it includes `FORCE_ASSERTIONS=1,LLVM_ASSERTIONS=1` in the make flags.

Let me know if that's not clear.

IanButterworth: Ah ok. Thanks

PR #245, opened 2022-11-14
[windows] Upload `.zip` files

staticfloat:
We need to serve not only `.tar.gz` files, but `.zip` files as well, as many windows users do not have an easy way to extract tarballs.


staticfloat:
This should solve issues like the one reported here: https://discourse.julialang.org/t/julia-nightly-is-27-days-old/90054

DilumAluthge (approved):
Is there a reason that we use PowerShell to create the `.zip` file, instead of using 7zip or something else that we already have lying around?

But LGTM regardless.

staticfloat:
Because if we stop shipping p7zip, this will keep on working.

DilumAluthge:
In hindsight we should have gotten this in before 1.8.3.

DilumAluthge:
At least it will be in 1.8.4 (once I backport this PR to the `release-julia-1.8` branch in this repo).

PR #246, opened 2022-11-14
Don't hard code the analyzegc deps

gbaraldi:
The hard coded makefile call exists so don't hard code it :)

EDIT: The reason for this is so I can pass mimalloc.h to analyzegc. Or any other dep for that matter.

### DIFF
--- a/pipelines/main/misc/analyzegc.yml
+++ b/pipelines/main/misc/analyzegc.yml
@@ -22,7 +22,7 @@ steps:
           make --output-sync -j$${JULIA_CPU_THREADS:?} -C src install-analysis-deps
           echo "+++ run clangsa/analyzegc"
           make --output-sync -j$${JULIA_CPU_THREADS:?} -C test/clangsa
-          make --output-sync -j$${JULIA_CPU_THREADS:?} -C src analyzegc
+          make --output-sync -j$${JULIA_CPU_THREADS:?} -C src analyze
         timeout_in_minutes: 60
         agents:
           queue: "julia"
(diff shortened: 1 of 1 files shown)


vchuravy (approved):
Seems fine to me, another minor annoyance is that  `analyzegc` should have a dependency on this target, but thats outside the scope of this.

pipelines/main/misc/analyzegc.yml
```diff
@@ -19,7 +19,7 @@ steps:
                 - "/cache/repos:/cache/repos"
         commands: |
           echo "--- Install in-tree LLVM dependencies"
-          make --output-sync -j$${JULIA_CPU_THREADS:?} -C deps install-llvm install-clang install-llvm-tools install-libuv install-utf8proc install-unwind
+          make --output-sync -j$${JULIA_CPU_THREADS:?} -C src install-analysis-deps
           echo "+++ run clangsa/analyzegc"
           make --output-sync -j$${JULIA_CPU_THREADS:?} -C test/clangsa
           make --output-sync -j$${JULIA_CPU_THREADS:?} -C src analyzegc
```
vtjnash, line 25: could you also make this use the more general name? (add in v1.9 by JuliaLang/julia#44900)
```suggestion
          make --output-sync -j$${JULIA_CPU_THREADS:?} -C src analyze
```

gbaraldi: 👍

gbaraldi:
Can we merge this :)?

Issue #239, opened 2022-11-07
Plan for OpenBLAS threads on CI


DilumAluthge:
1. In the `test_julia.sh` script, set `OPENBLAS_NUM_THREADS=1` before we start the Julia test suite.
2. At the beginning of the desired LinearAlgebra test sets, call `LinearAlgebra.BLAS.set_num_threads(4)` to make sure that we test the multithreading code paths.
3. In the `testdefs.jl` code in JuliaLang/julia, after running each test set, call `LinearAlgebra.BLAS.set_num_threads(1)`.

gbaraldi:
Do we want to make this a windows only thing? I'm not sure it's worth losing the coverage of the other OSs and they don't seem to have any issue

DilumAluthge:
> Do we want to make this a windows only thing?

Sounds good to me!

PR #247, opened 2022-11-14
Set the number of BLAS threads to 2.

gbaraldi:
As an experiment, I set the blas threads to 1, and enable threads on the matmul testset.

Ref #239 

Edit:
Set BLAS threads to 2 to reduce change of OOM on windows and reduce contention

### DIFF
--- a/.buildkite/hooks/post-checkout
+++ b/.buildkite/hooks/post-checkout
@@ -3,9 +3,9 @@
 set -euo pipefail
 
 # Buildkite users can trigger new builds with these values overridden to
-# test specific julia gitshas 
-UPSTREAM_URL="${UPSTREAM_URL:-https://github.com/gbaraldi/julia.git}"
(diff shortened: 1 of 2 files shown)


staticfloat:
Discussed this on the call, and we decided to set the number of threads to 2 everywhere.  That should solve our memory issues, and also exercise multithreading paths.

gbaraldi:
Is this good to merge?

DilumAluthge:
Can you update the PR title to reflect the final contents of the PR?

staticfloat:
Don't merge this yet; I think Kristoffer and I have found that we are actually only supposed to be using one thread on CI.  I think we need to solve the timing issue in Distributed's setting of threads, first.

gbaraldi:
So distributed should be passing the number of threads as 1 already? But doesn't?

staticfloat:
https://github.com/JuliaLang/julia/pull/46844#issuecomment-1315790913

PR #252, opened 2022-11-15
Make sure that `make full-source-dist` doesn't regress

DilumAluthge:
In a follow-up PR, we'll take this output and upload it to S3. But for now, let's just get this PR in, so that we can make sure that the full source dist doesn't regress again.


DilumAluthge:
I'm getting:

```
Cannot change ownership to uid 1000, gid 1000: Invalid argument
```

@staticfloat @ararslan Any ideas what's going on?

pipelines/main/misc/fullsrcdist.yml
```diff
@@ -0,0 +1,33 @@
...
+          - staticfloat/sandbox#v1:
+              rootfs_url: https://github.com/JuliaCI/rootfs-images/releases/download/v4.8/package_linux.x86_64.tar.gz
+              rootfs_treehash: "2a058481b567f0e91b9aa3ce4ad4f09e6419355a"
+              workspaces:
+                - "/cache/repos:/cache/repos"
+          - JuliaCI/julia#v1:
+              # Drop default "registries" directory, so it is not persisted from execution to execution
+              persist_depot_dirs: packages,artifacts,compiled
+              version: '1.6'
+        timeout_in_minutes: 10
```
ararslan, line 25: I think `full-source-dist` may take more than 10 minutes... I've never timed it but anecdotally it feels like an eternity. It does a source build of all dependencies then Julia then all documentation.

pipelines/main/misc/fullsrcdist.yml
```diff
@@ -0,0 +1,33 @@
...
+        commands: |
+          make full-source-dist USE_BINARYBUILDER=0
```
ararslan, line 27:
```suggestion
          make full-source-dist USE_BINARYBUILDER=0
          make light-source-dist USE_BINARYBUILDER=0
```
This would be convenient so that we can upload as release assets on GitHub.

pipelines/main/misc/fullsrcdist.yml
DilumAluthge: Wait, it actually builds Julia? I thought it just downloads the source code of the dependencies.

ararslan:
> Any ideas what's going on?

None at all from me. Where are you seeing that? I still haven't gotten used to navigating the buildkite UI...

pipelines/main/misc/fullsrcdist.yml
ararslan: Yeah haha. It builds Julia so that it can use the built binary to build the documentation.

DilumAluthge: 😬

DilumAluthge: Hmmm. A full from-source build of the dependencies and Julia takes more than 30 minutes on our fastest CI machines. See e.g. https://buildkite.com/julialang/julia-master-scheduled/builds/159#01847a53-374d-441c-90cc-c4dcfef6ae22

Maybe we should move the `full-source-dist` job to the once-daily scheduled pipeline. That's where the existing `USE_BINARYBUILDER=0` job lives.

DilumAluthge: Anyway, let's move this to the once-daily scheduled pipeline (https://buildkite.com/julialang/julia-master-scheduled).

That pipeline runs once a day, so if there's a regression on Julia master, we catch it pretty quickly. But we don't run it on PRs, so we save some CI time that way.

staticfloat: We should use the same trick that we use elsewhere for the documentation; download a prebuilt Julia, unpack it in the correct place.  Then this won't rebuild Julia because it will already exist.

staticfloat:
Try adding `uid` and `gid` parameters to the sandbox plugin, like this: https://github.com/JuliaCI/julia-buildkite/blob/709d22629922ec7d10d2213158f4c999f34968d5/pipelines/main/platforms/build_linux.yml#L17-L18

pipelines/main/misc/fullsrcdist.yml
DilumAluthge: I'm now downloading a pre-built Julia, but it seems to insist on building everything (including LLVM) from source.

DilumAluthge: @ararslan What's the difference between `make full-source-dist` and `make full-source-dist USE_BINARYBUILDER=0`? Shouldn't they both produce the same output? Presumably we shouldn't have to build LLVM from source just to be able to build the docs and download the source tarballs of all dependencies, right?

ararslan: `USE_BINARYBUILDER=0` ensures that the source code of the dependencies is included in the resulting tarball alongside Julia's source code. It's meant to be platform-agnostic and buildable without network access, which IIUC isn't possible with the default JLL-based dependencies.

DilumAluthge: Sure, but shouldn't `make full-source-dist` do that regardless of the value of `USE_BINARYBUILDER`?

I guess what I'm getting at is, do we need to actually do a full from-source build of LLVM etc just to build this tarball? AFAICT, the tarball needs the following contents:
1. Source code of Julia
2. Source code of all of Julia's dependencies
3. HTML documentation

It seems to me that we can use a regular Julia build to build the HTML docs, and then download the source code of all the dependencies, and then stick everything together.

ararslan: We definitely could do that, the Makefile just isn't set up that way

DilumAluthge:
I'm closing this to "un-lick the cookie". Maybe someone else can pick up this PR and resurrect it.

PR #253, opened 2022-11-16
On tags, skip the upload if the files already exist on S3

DilumAluthge:
# Description

If `${TAR_VERSION}` does NOT match the regex `^\d*?\.`, then we always upload.

If `${TAR_VERSION}` DOES match the regex, then we do a HEAD request to see if the files already exist. If they do exist, we don't upload. If they don't exist, we do upload.

# Motivation

I'd like to be able to trigger new builds without worrying about overwriting existing tagged binaries on S3.

This is only relevant for tags.

### DIFF
--- a/utilities/build_envs.sh
+++ b/utilities/build_envs.sh
@@ -108,37 +108,40 @@ S3_BUCKET="${S3_BUCKET:-julialangnightlies}"
 S3_BUCKET_PREFIX="${S3_BUCKET_PREFIX:-bin}"
 
 # We generally upload to multiple upload targets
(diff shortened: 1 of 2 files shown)


DilumAluthge:
The treehashes need to be re-signed. However, I'll wait to re-sign them until after we've finished reviewing this PR and making any necessary changes.

DilumAluthge:
In theory, we're vulnerable to TOCTOU here.

But we only start the job once we are holding the concurrency group lock, so we should be good, right?

staticfloat:
Why are we shy about replacing tagged binaries?  We're not uploading directly to the `julialang2` bucket, so I feel it's totally okay for us to overwrite them.  This feels like a lot of complexity for unclear benefit.

PR #261, opened 2022-11-29
Backports for Julia 1.8

DilumAluthge:
- [x] #249
    - 91821e5576ee4fd67f86cc78582e5985fdd2ec1a
- [x] #263
- [x] Sign treehashes


DilumAluthge:
I somehow messed up this backport. It didn't actually get the "eager" commit, it only got the "sign treehashes" commit.

DilumAluthge:
I went into the repository settings and disabled the "Allow merge commits" and "Allow rebase merging" options. So the only allowed option is the "Allow squash merging" option. So now backporting should be easier, because I'll only need to cherry-pick the single squashed commit.

DilumAluthge:
Ohhhh. @staticfloat is that why you chose to rebase merge #249? Because it allows me to cherry-pick the useful commits but not cherry-pick the "sign treehashes" commits, and thus you don't get merge conflicts when cherry-picking?

staticfloat:
Yeah, in general, I prefer to retain the commit structure of the original PR; if the author kept things as three separate commits, it makes sense to me to maintain that structure as well.

PR #262, opened 2022-11-29
Use p7zip to create our windows zipballs

staticfloat:
It turns out powershell embeds backslashes as path separators, which chokes some unzip algorithms, because of course it does.  Let's use p7zip instead.

X-ref: https://github.com/icsharpcode/SharpZipLib/issues/617
X-ref: https://buildkite.com/julialang/julia-buildkite-plugin/builds/100#0184c19d-5e4a-42ed-b13e-94b0213e2a8e


DilumAluthge:
Of course it does.

DilumAluthge:
We should also backport this PR to the `release-julia-1.8` branch in this repo.

Issue #243, opened 2022-11-10
Run only a subset of tests on PRs that only modify the manual


DilumAluthge:
If a PR only modifies the manual, then we only need to run a subset of jobs, including:
1. Build a regular (non-assert) build of Julia on x86_64-linux-gnu
2. Build the docs
3. Run the doctests

staticfloat:
I think the way to do this is to have a "filter" that looks at the diff, and skips launching a job if the diff does not contain changes to a particular group of files.  We can also construct this negatively, e.g. filter out elements of a diff, and then run if any diff remains.

The way I'm imagining this is to write a buildkite plugin that does something like the following:

```
steps:
  - label: launch jobs
    plugins:
      - JuliaCI/diff-filter#v1:
          docs:
            # Only run docs if we have a diff that changes something in `doc/`
            mode: positive
            patterns:
              - "doc/**"
          build:
            # Run build if anything changes, excluding changes to docs, README, etc...
            mode: negative
            patterns:
              - "doc/**"
              - "README.md"
    commands: |
      if [[ $${BUILDKITE_PLUGIN_DIFF_FILTER_TRIGGERED_DOCS} == 1 ]]; then
          buildkite-agent upload docs.yml
      fi
      if [[ $${BUILDKITE_PLUGIN_DIFF_FILTER_TRIGGERED_BUILD} == 1 ]]; then
          buildkite-agent upload build.yml
      fi
```

staticfloat:
To filter the build, we can use this code: https://github.com/staticfloat/forerunner-buildkite-plugin/blob/856c0f5d49a5d2d84270cb14b41cbb4fa5d3dc8b/hooks/command#L8-L9 and https://github.com/staticfloat/forerunner-buildkite-plugin/blob/856c0f5d49a5d2d84270cb14b41cbb4fa5d3dc8b/hooks/command#L59-L100

giordano:
In Yggdrasil we collect the files modified in a PR/commit: https://github.com/JuliaPackaging/Yggdrasil/blob/fb8152880d5bb011c4ab2a23b84fff2327c1ee17/azure-pipelines.yml#L81-L86.  Valentin will have done something similar with buildkite in https://github.com/JuliaPackaging/Yggdrasil/pull/4771

giordano:
[...]

PR #264, opened 2022-12-01
Linux: on every PR, run `make -C deps USE_BINARYBUILDER=0` and make sure that the working directory is clean

DilumAluthge:
This lets us catch checksum issues a little bit earlier, so that we don't have to wait for the scheduled from-source build.

### DIFF
--- a/pipelines/main/launch_unsigned_jobs.yml
+++ b/pipelines/main/launch_unsigned_jobs.yml
@@ -60,10 +60,18 @@ steps:
           buildkite-agent pipeline upload .buildkite/pipelines/main/misc/pdf_docs/build_pdf_docs.yml
           buildkite-agent pipeline upload .buildkite/pipelines/main/misc/embedding.yml
           buildkite-agent pipeline upload .buildkite/pipelines/main/misc/llvmpasses.yml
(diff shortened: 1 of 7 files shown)


DilumAluthge:
This will fail until https://github.com/JuliaLang/julia/pull/47763 is merged.

staticfloat:
Yes, just rerun the check…. But IMO we shouldn’t run this on every job. That will hit upstream servers needlessly. We should only run it if a checksum file has been changed.

DilumAluthge:
We've talked a lot about having jobs that are only triggered when certain files are in the PR diff. I think this job could be a nice guinea pig for us to try that out?

I guess we basically need to get the PR diff, and then if anything under `deps` is modified, we launch this job, but otherwise, we don't launch it.

How do we get the PR diff? If we get it from the GitHub API, then we'll need to use a GitHub personal access token to avoid being rate-limited, and thus we'll need to make the "PR diff" job a signed job.

giordano:
> How do we get the PR diff?

https://github.com/JuliaCI/julia-buildkite/issues/243#issuecomment-1310785115

PR #267, opened 2022-12-16
Update sandbox-buildkite-plugin instead of locking to v1.2

staticfloat:
This should include a fix meant to solve `EBUSY` errors due to copying `/bin/bash` instead of moving it.


DilumAluthge (approved):

staticfloat:
Looks good: https://buildkite.com/julialang/julia-buildkite/builds/1206#01851c2e-a980-4887-bd01-9795fdc628f8/608-609

PR #268, opened 2022-12-20
Only run doctests on `doctest_capable` agents

staticfloat:
We're going to dodge the issue of slightly-varying floating point results on different machines by only running doctests on amdci machines, which self-report by setting `doctest_capable`.


staticfloat:
X-ref: https://github.com/JuliaDocs/Documenter.jl/issues/1999

DilumAluthge (approved):
It would be nice to fix the doctests in the long term, but at least this will tide us over in the short term.

PR #269, opened 2022-12-21
[release-1.6 only] Drop `--ci` argument to test suite

staticfloat:
We don't have `--ci` (or `--force-net`, its replacement) in v1.6


DilumAluthge:
If necessary, we could backport `--force-net` to Julia 1.6.

PR #275, opened 2022-12-29
Update sandbox-buildkite-plugin instead of locking to v1.2

giordano:
Same as #267, but on the `release-1.8` branch.


giordano:
Ah, that's already in #274.

PR #277, opened 2023-01-13
Rebase Ian's fix on top of new main and try to rerun it.

kshyatt:


kshyatt:
OK I think we are *finally* ready to review. It looks like the `base/` dir at the main Julia codecov repo was populated properly with line by line info!

pipelines/scheduled/coverage/coverage.yml
```diff
@@ -5,7 +5,8 @@ steps:
         key:   "coverage"
         # Comment out the following `if:` statement for testing.
         # Please remember to uncomment it before merging the PR.
-        if: pipeline.slug == "julia-master-scheduled"
+        # TODO: uncomment before merge
+        # if: pipeline.slug == "julia-master-scheduled"
```
staticfloat, line 9: Should this be uncommented?

kshyatt:
I think it is, above. Should we delete the TODO and the following line?

staticfloat:
I don't think it is.

PR #281, opened 2023-02-04
Exclude all coverage files that aren't in `base/` or `stdlib/`

kshyatt:
Should get rid of the extraneous `Foo.jl`s etc.


kshyatt:
Test fail again seems unrelated

PR #287, opened 2023-03-06
Add multithreaded image environment variables

pchintalapudi:


utilities/build_envs.sh
```diff
@@ -82,6 +82,8 @@ fi
 JULIA_CPU_TARGET="$(printf ";%s" "${JULIA_CPU_TARGETS[@]}")"
 export JULIA_CPU_TARGET="${JULIA_CPU_TARGET:1}"
 
+export JULIA_IMAGE_THREADS="$JULIA_CPU_THREADS"
+export JULIA_IMAGE_TIMINGS=1
```
staticfloat, line 86: I think this is too noisy

staticfloat, line 86:
```suggestion
```

PR #288, opened 2023-03-09
Search for 7z in multiple places

staticfloat:
7zip's install location changed recently; let's use `PATH` and search for it like a normal user.


DilumAluthge (approved):

staticfloat:
We could; either way is fine with me.

PR #292, opened 2023-03-31
update macos stdlib pkgimage checksums after codesigning

IanButterworth:
Ref. https://github.com/JuliaLang/julia/pull/49211


IanButterworth:
@staticfloat I don't see this being called in `build aarch64-apple-darwin` but I do see the stdlibs being codesigned there.

I believe it needs to happen there otherwise the `test` workers won't be able to load the stdlib caches

IanButterworth:
It seems like this fix needs to happen between here

https://github.com/IanButterworth/julia/blob/94a78af6db66f7204fbb58ca11a02d0502d6476d/Makefile#L507-L509

IanButterworth:
I'm giving that a go here https://github.com/JuliaLang/julia/pull/49132

staticfloat:
> I don't see this being called in build aarch64-apple-darwin but I do see the stdlibs being codesigned there.

That codesigning is worthless, as the build step doesn't have access to codesigning keys, so they're all just ad-hoc signed (which happens anyway when you link with `ld` on newer macOS machines).  I have a lingering PR to remove those steps from the build process: https://github.com/JuliaLang/julia/pull/48545

The test workers re-sign everything when they download them, right here: https://github.com/JuliaCI/julia-buildkite/blob/c5449d1ca3ede5752654898caa923ebc32f9a55d/utilities/test_julia.sh#L33.  You can see the outputs in the test job of this PR right here: https://buildkite.com/julialang/julia-buildkite/builds/1296#0187394f-53cc-4067-959a-0bce9ff5fb5e/492-575

IanButterworth:
But that codesigning is changing the files, meaning the test runners can't load them, so either the codesigning needs to not be done (if worthless) or the checksum needs updating there for tests to pass.

So it sounds like right now this needs to be called in both places?

IanButterworth:
Seems to have worked? Just waiting for a x86 MacOS runner

IanButterworth:
Both macos runners passed. Windows failed tests which I guess is unrelated. i retried.

Also the .dmg uploaded works locally

staticfloat:
Looks good to me!

PR #294, opened 2023-04-05
Add builder and tester for PowerPC with asserts; ppc64le-linux-gnuassert

ahsansaghir:
This PR adds pipeline for builder and tester for PowerPC with asserts enabled (for both Julia and LLVM).

### DIFF
--- a/pipelines/main/platforms/build_linux.arches
+++ b/pipelines/main/platforms/build_linux.arches
@@ -4,6 +4,7 @@ package_linux          x86_64-linux-gnu               x86_64         x86_64     
 package_linux          x86_64-linux-gnuassert         x86_64         x86_64         FORCE_ASSERTIONS=1,LLVM_ASSERTIONS=1,JL_CFLAGS=-Werror,JL_CXXFLAGS=-Werror    .             v5.26         5b0b851aca3c941b900a1301c13922c6cfc7f211
 package_linux          aarch64-linux-gnu              aarch64        aarch64        .                                                                             .             v5.26         dcf39d10ba43bf13c75d5031c3a88f125780033b
 package_linux          powerpc64le-linux-gnu          powerpc64le    powerpc64le    .                                                                             .             v5.26         45378b056347797ff6a5411c05f86a9119d17d19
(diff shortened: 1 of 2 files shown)


staticfloat:
I think `-Werror` might be a little bit ambitious.  ;)

If you want to experiment with your own branch of Julia, you can (temporarily) modify [these lines](https://github.com/JuliaCI/julia-buildkite/blob/bc92933c788311c33204b505a459932b15462be9/.buildkite/hooks/post-checkout#L7-L8) to fetch your own version of Julia and run the tests on this PR; that will allow you to iterate on any changes needed in Julia while modifying this code.

If you're going to be iterating a lot, please comment out the other build jobs so that you're only building ppc64le over and over again.  :)

ahsansaghir:
> I think `-Werror` might be a little bit ambitious. ;)
> 
> If you want to experiment with your own branch of Julia, you can (temporarily) modify [these lines](https://github.com/JuliaCI/julia-buildkite/blob/bc92933c788311c33204b505a459932b15462be9/.buildkite/hooks/post-checkout#L7-L8) to fetch your own version of Julia and run the tests on this PR; that will allow you to iterate on any changes needed in Julia while modifying this code.
> 
> If you're going to be iterating a lot, please comment out the other build jobs so that you're only building ppc64le over and over again. :)

Yeah, I did not realize that `-Werror` would be ambitious. :). Thanks a lot for the tips! I will iterate over without `-Werror` and see how it goes. :)

ahsansaghir:
@staticfloat This PR is now ready for review. Regarding `rr`; I discussed with Valentin and it seems like it would be very hard to get `rr` working on PPC. It requires the use of precise performance counters and apparently the ones that are needed aren't precise enough on PPC9.

I have divided formatting and actual changes into two commits to make it easier to compare the actual changes.

ahsansaghir:
FYI @vchuravy. This is the PR that I have put up for adding the builder/tester for PowerPC with asserts enabled.

staticfloat (approved):
This looks good to me!  Thanks for pushing this through, @ahsansaghir!

Do you think that these assert builds will be somewhat temporary, (as we intend to crush the PPC-specific bugs in the near term) or do you think we should keep them forever?

vchuravy:
I think we should keep them permanently, I think we undertest non x86,

PR #295, opened 2023-04-08
`upload_macOS` needs to run on the same architecture as the build job

vtjnash:


staticfloat (approved):

DilumAluthge:
We need to fix the other upload jobs too. There are multiple `upload_macOS` files.

staticfloat:
I think this already updates all of them?

PR #296, opened 2023-04-21
Disable `-Wunused-but-set-variable` on `aarch64-apple-darwin`

staticfloat:


giordano:
If there's only a warning, how about disabling only that?

DilumAluthge:
Do we need to change the format of the `MAKE_FLAGS` column? Right now, the entries in `MAKE_FLAGS` are comma-separated, but obviously this will be a problem if the entries themselves contain commas.

staticfloat:
Actually, I don’t want a comma I want a space. I just havent figured out how to do it yet. :P

giordano:
I presume you already tried quotes?

staticfloat:
I struggled for about half an hour trying to shoe-horn multiple values into the `JL_CFLAGS` make flag here, and couldn't do it.  I'm just going to merge a lack of `-Werror` and we can improve things later.

PR #302, opened 2023-05-11
Add profiling-enabled (ITTAPI + Tracy) Linux build

topolarity:
This just adds a builder for now, and Linux-only.

This is my first time modifying our buildkite pipelines, so let me know if any extra action is needed from me re: testing etc.


pchintalapudi:
Can we also add WITH_TIMING_COUNTS to this build? That would potentially allow us to track various summary metrics over the history of the master branch.

topolarity:
@staticfloat This build is now passing - I think all we need is review and yaml signing

PR #307, opened 2023-06-14
Use new notarytool on macos

gbaraldi:
According to apple the `--wait` flag is enough to supercede the loop we had.

Also https://github.com/thezik/notarizer for reference


staticfloat:
This looks like it's passing!  I had to rebase and resign, so let's see if it goes through again.

PR #308, opened 2023-06-20
Don't multiversion on 32bit

gbaraldi:
We run out of memory when multiversioning on 32 bit. Target only sandybridge which should be a reasonable default anyway.


utilities/build_envs.sh
```diff
@@ -32,10 +32,8 @@ case "${ARCH?}" in
         ;;
     i686)
         JULIA_CPU_TARGETS+=(
-            # We require SSE2, etc.. so `pentium4` is our base i686 feature set
-            "pentium4"
-            # Add sandybridge level similar to x86_64 above
-            "sandybridge,-xsaveopt,clone_all"
+            # Target only sandybridge because we run out of memory when multiversioning
+             "sandybridge"
```
vchuravy, line 36: Uhm... I think technically we must support pentium4.

vchuravy: Or it would at least be a bigger decision than I thought.

gbaraldi: Why must we support it?

staticfloat: I agree with Valentin.  Unless we're willing to officially deprecate things, we must error on the side of caution.  Let's throw away `sandbridge` and keep `pentium4` for now, until we decide to jump the gun to sandybridge.  For the record, sandybridge was released "only" 12 years ago, so it's not outside of the realm of possibility that computers built 10 years ago had pre-sandybridge chipsets.

I have an alternative idea on how to help this; I don't think it's multiversioning that's crashing, I think it's LLVM image generation, which we're multithreading by default.  I'll try limiting that to one thread and seeing if it helps.

gbaraldi: We already don't multithread, not multiversioning does solve it for me, so it does affect stuff.

staticfloat: > We already don't multithread

What makes you say that?  It looks to me like [we set `JULIA_IMAGE_THREADS` to `JULIA_CPU_THREADS` unconditionally](https://github.com/JuliaCI/julia-buildkite/blob/61513b36584fb3b06626e9c246b42b5ec4f0cc61/utilities/build_envs.sh#L85).

gbaraldi: Right, but IIRC the Julia code has a 32 bit define for this.

staticfloat: I was unaware of this, you're correct: https://github.com/JuliaLang/julia/blob/a8ef87369749abc16e334014108f3080b76ab507/src/aotcompile.cpp#L1407-L1410

staticfloat:
I've adjusted to only `pentium4`, with a descriptive comment.

PR #309, opened 2023-06-20
Don't multithread image generation on 32-bit architectures

staticfloat:
We hit OOM errors pretty often from LLVM-land, let's see if this helps.


staticfloat:
Gabriel points out that this shouldn't help: https://github.com/JuliaLang/julia/blob/a8ef87369749abc16e334014108f3080b76ab507/src/aotcompile.cpp#L1407-L1410

staticfloat:
As expected, this doesn't help.

PR #313, opened 2023-07-29
Upgrade builder rootfs images to v6.00

staticfloat:
This bumps the minimum glibc version to v2.17+ on `i686` and `x86_64`.


vtjnash (commented):
🎉

PR #317, opened 2023-11-01
prioritize release branches

IanButterworth:
This is a bit of a guess.

The example on https://buildkite.com/docs/pipelines/managing-priorities#example seems more convoluted. This tries something simpler..


DilumAluthge:
Does Buildkite interpret `build.branch =~ /^release-/` as a boolean, or an integer?

DilumAluthge:
Because I'm seeing the following in the logs:

```
fatal: Failed to upload and process pipeline: Pipeline upload rejected: One of the steps you provided was invalid: is not a number
```

fredrikekre:
Yea probably a bool. In any case, I think we should do something similar to what they do in their example anyway, because right now we set the priority on the pipeline that launches jobs but we also need to prioritize the jobs that these launch.

PR #322, opened 2023-11-14
Update `rr` from 5.5 to 5.7

DilumAluthge:


staticfloat (approved):

vtjnash:
The main fixes we need require 5.7 (https://github.com/rr-debugger/rr/releases/tag/5.7.0)

DilumAluthge:
Okay, this PR now updates to rr 5.7.

vtjnash:
Not great that rr functionality was the test that failed, but I think we see that failure commonly already?

DilumAluthge:
I'll retry the job.

DilumAluthge:
Alright, it took a couple of retries, but finally it passed.

@Keno @vtjnash In one of the failed jobs, I saw this:

```
`rr` returned 1, packing and uploading traces...
 rr pack [OPTION]... [<trace-dir>]
  --symlink                  Create symlinks to all mmapped files
                             instead of copying them.
Eliminates duplicate files in the trace directory, and copies files into
the trace directory as necessary to ensure that all needed files are in
the trace directory and none of them are links to files outside the
trace directory. This makes the trace directory independent of changes
to other files and ready to be transported elsewhere (e.g. by packaging
it into a ZIP or tar archive).
Be careful sharing traces with others; they may contain sensitive information.
Global options:
  --disable-cpuid-faulting   disable use of CPUID faulting
  --disable-ptrace-exit_events disable use of PTRACE_EVENT_EXIT
  --resource-path=PATH       specify the paths that rr should use to find
```

This makes it seem like we are invoking `rr pack` in an incorrect way, and thus getting a help message.

DilumAluthge:
Link to the job with the incorrect `rr pack` invocation: https://buildkite.com/julialang/julia-buildkite/builds/1697#01916694-1572-42df-8b68-31d12741023a

DilumAluthge:
I'm guessing the problematic line is here:

https://github.com/JuliaCI/julia-buildkite/blob/890b81c85752123f0097f8dce374425379523a5e/utilities/rr/rr_capture.jl#L116-L116

PR #329, opened 2023-12-06
Remove Win32

vchuravy:
Win 32bit is ~1% of Win 64bit according to the pkg statistics.

Generally speaking it consumes an outsized amount of developer resources
trying to keep everything working on it. I would thus like to stop supporting
Win 32bit as of Julia 1.11


inkydragon:
I'd like to say this is a big change, since we drop a Tier 1 platform.
Maybe it's better to discus it in https://github.com/JuliaLang/julia or discourse first.

Or we can change Win32 from Tier1 to Tier3, add a news, and wait for several months, before we remnove it.

I'm all in favor of removing win32 support, since MSYS2, the compilation environment currently used by CI, no longer provides a new win32 installer [since 2020-05-22](https://github.com/msys2/msys2-installer/releases/tag/2020-05-22).

https://discourse.julialang.org/t/is-anyone-using-32-bit-julia-on-windows/89525

PR #330, opened 2023-12-09
run SIGINFO profile before timeout-terminating test runner

IanButterworth:
It'd be helpful to know where timing-out processes are in julia code

### DIFF
--- a/utilities/timeout.jl
+++ b/utilities/timeout.jl
@@ -64,12 +64,17 @@ timer_task = @async begin
             println(stderr, "\n\nProcess failed to exit within $(term_timeout)s, requesting termination (SIGTERM) of PID $(proc_pid).")
         else
             println(stderr, "\n\nProcess failed to exit within $(term_timeout)s, requesting profile then termination (SIGTERM) of PID $(proc_pid).")
-            if Sys.islinux()
-                kill(proc, 30) # SIGUSR1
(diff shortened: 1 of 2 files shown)


utilities/timeout.jl
```diff
@@ -60,7 +60,19 @@ timer_task = @async begin
 
     # If the process is still running, ask it nicely to terminate
     if isopen(proc)
-        println(stderr, "\n\nProcess failed to exit within $(term_timeout)s, requesting termination (SIGTERM) of PID $(proc_pid).")
+        if Sys.iswindows()
+            println(stderr, "\n\nProcess failed to exit within $(term_timeout)s, requesting termination (SIGTERM) of PID $(proc_pid).")
+        else
+            println(stderr, "\n\nProcess failed to exit within $(term_timeout)s, requesting profile then termination (SIGTERM) of PID $(proc_pid).")
+            if Sys.islinux()
+                kill(proc, 10) # SIGUSR1
```
IanButterworth, line 68:
```suggestion
                kill(proc, 30) # SIGUSR1
```

IanButterworth:
I'm guessing rr is somehow getting in the way of the SIGUSR1, as I can't get the profile to trigger.

Also, this will only be useful if the parent process is running the hanging task, which it most likely isn't.

PR #333, opened 2024-02-01
Update coverage test timeout

vtjnash:
This is usually run on a 128 core machine, and gets about 90% finished before getting killed every night


DilumAluthge (approved):

staticfloat:
asan timed out, but we’re calling this good

PR #341, opened 2024-02-26
Improvements to modern machine coverage in JULIA_CPU_TARGET

pthariensflame:
This is primarily motivated by use cases on Apple Silicon machines, but tries to also cover other modern developments not handled by existing target strings. Only addresses x86_64 and AArch64 needs.

Redo of JuliaCI/julia-buildbot#282.


utilities/build_envs.sh
```diff
@@ -24,10 +24,14 @@ case "${ARCH?}" in
         JULIA_CPU_TARGETS+=(
             # Absolute base x86_64 feature set
             "generic"
+            # Architecture implemented by Apple's Rosetta 2 translation layer
+            "westmere,aes"
```
vchuravy, line 28: I am not convinced that we should add this. People should just be using the native binaries at this point.

utilities/build_envs.sh
```diff
@@ -24,10 +24,14 @@ case "${ARCH?}" in
...
             # Add sandybridge level (without xsaveopt) and that clones all functions
             "sandybridge,-xsaveopt,clone_all"
             # Add haswell level (without rdrnd) that is a diff of the sandybridge level
-            "haswell,-rdrnd,base(1)"
+            "haswell,-rdrnd,base(2)"
+            # A common baseline for modern x86-64 server CPUs
+            "x86-64-v4,base(2)"
```
vchuravy, line 34:
```suggestion
            "x86-64-v4,-rdrnd,base(2)"
```

utilities/build_envs.sh
pthariensflame: Unfortunately, that’s not always possible due to other dependencies or (as in my own case) due to needing to test packages under Rosetta 2 as a deliberate use case.  I can currently get away with using the existing x86_64 builds but need to constantly specify the target string in both the environment variable and command-line flag forms; omitting either one causes crashing with either illegal instructions or internal errors in the Julia binary…and inspecting the package dylibs with elfx86exts shows me there’s a lot of performance being left on the table.

vchuravy: > omitting either one causes crashing with either illegal instructions or internal errors in the Julia binary

That would be worth bug reports, but new targets take up significant space and and this needs to be broadly useful.

pthariensflame: @vchuravy Understood; I'll remove it.

staticfloat (approved):
LGTM, but let's have @vtjnash take a look before we merge.

vtjnash:
What does this add for amd64 machines? Usually there needs to be some major feature difference for it to generate any code. Anyways, I am not the one to ask about this

pthariensflame:
@vtjnash 
> What does this add for amd64 machines? Usually there needs to be some major feature difference for it to generate any code. Anyways, I am not the one to ask about this

Basically, AVX512 support.  Haswell represents the baseline form of x86-64-v3; you can see the differences from there in the table here: https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

PR #345, opened 2024-03-26
Add building and testing of PGO+LTO LLVM and Julia

Zentrik:
I wasn't sure how to test this locally, so hopefully it works.
I ended up just using the profile from Julia being built for PGO, locally it seems to work fine. If anyone has better ideas for what to profile, please do say.

I'm not sure if this will timeout but I wasn't sure how to adjust the timeout either.
I don't think we need to run the full test suite but it seems good to run at least some stuff to make sure nothing got messed up. Also, I don't want to slow down getting a green checkmark on CI so maybe the tests should be allowed to fail?


Zentrik:
I don't think I have permissions to fix `ERROR: Pipeline '.buildkite/pipelines/main/launch_upload_jobs.yml' fails treehash signature check!  You may need to re-run cryptic/bin/sign_treehashes!`.

Zentrik:
Build currently failing as the bump to patchelf https://github.com/JuliaLang/julia/pull/53845 introduced a check that the compiler supports C++ 17. Not sure why that check is failing on CI as it does seem to be using clang 16.

Zentrik:
Maybe `clang` is picking the wrong `gcc`?

Zentrik:
Ok so the actual error is 
```
configure:6320: /cache/build/builder-amdci4-6/julialang/julia-buildkite/contrib/pgo-lto/stage0.build/usr/tools/clang++ -m64  -std=c++17 -c -fprofile-generate=/cache/build/builder-amdci4-6/julialang/julia-buildkite/contrib/pgo-lto/profiles -Xclang -mllvm -Xclang -vp-counters-per-site=6  conftest.cpp >&5
conftest.cpp:437:10: fatal error: 'initializer_list' file not found
#include <initializer_list>
```
https://github.com/JuliaLang/julia/pull/45641#issuecomment-1168879156 mentions something about libstdc++ detection, so maybe something to do with that.

haampie:
Is that a partial gcc 10 install? No g++ 10?

You can force clang to use gcc 9 if necesary with `--gcc-install-dir=/usr/lib/gcc/x86_64-linux-gnu/9`

Zentrik:
I'm quite out of my depth here but I'm not sure gcc is installed in `/usr/lib/gcc/x86_64-linux-gnu/9` because this was printed which seems to suggest it's installed elsewhere 
```
COLLECT_LTO_WRAPPER=/usr/local/bin/../libexec/gcc/x86_64-linux-gnu/9.1.0/lto-wrapper
Target: x86_64-linux-gnu
Configured with: /workspace/srcdir/gcc-9.1.0/configure --prefix=/workspace/destdir --with-build-sysroot=/workspace/destdir/x86_64-linux-gnu --with-sysroot=/workspace/destdir/x86_64-linux-gnu --with-gxx-include-dir=/workspace/destdir/x86_64-linux-gnu/include/c++/9.1.0 --target=x86_64-linux-gnu --host=x86_64-linux-gnu --build=x86_64-linux-gnu --disable-multilib --disable-werror --disable-bootstrap --enable-threads=posix --program-prefix=x86_64-linux-gnu- --with-arch=x86-64 --enable-languages=c,c++
```

Zentrik:
It does seem to find the correct gcc version now. Yes, it does seem to get through patchelf's configure phase but now fails at making it
```
/cache/build/builder-amdci4-0/julialang/julia-buildkite/contrib/pgo-lto/stage0.build/usr/tools/clang++ -DPACKAGE_NAME=\"patchelf\" -DPACKAGE_TARNAME=\"patchelf\" -DPACKAGE_VERSION=\"0.18.0\" -DPACKAGE_STRING=\"patchelf\ 0.18.0\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"patchelf\" -DVERSION=\"0.18.0\" -DHAVE_CXX17=1 -I. -I/cache/build/builder-amdci4-0/julialang/julia-buildkite/deps/srccache/patchelf-0.18.0/src    -Wall -Wextra -Wcast-qual -std=c++17 -D_FILE_OFFSET_BITS=64     -fprofile-generate=/cache/build/builder-amdci4-0/julialang/julia-buildkite/contrib/pgo-lto/profiles -Xclang -mllvm -Xclang -vp-counters-per-site=6 -MT patchelf.o -MD -MP -MF .deps/patchelf.Tpo -c -o patchelf.o /cache/build/builder-amdci4-0/julialang/julia-buildkite/deps/srccache/patchelf-0.18.0/src/patchelf.cc
/cache/build/builder-amdci4-0/julialang/julia-buildkite/deps/srccache/patchelf-0.18.0/src/patchelf.cc:19:10: fatal error: 'algorithm' file not found
#include <algorithm>
```

haampie:
The build environment likely contains _just_ the C compiler part of GCC 10, and clang has poor detection & heuristics for the GCC install dir. If multiple are in the same prefix it picks the newest, and it will do so also when there are missing components like C++ stdlib header files.

Use `--gcc-install-dir=/usr/lib/gcc/x86_64-linux-gnu/9`

Zentrik:
```
/cache/build/builder-amdci5-3/julialang/julia-buildkite/contrib/pgo-lto/stage0.build/usr/tools/clang  -fprofile-generate=/cache/build/builder-amdci5-3/julialang/julia-buildkite/contrib/pgo-lto/profiles -Xclang -mllvm -Xclang -vp-counters-per-site=6 -O2 -m64 --gcc-install-dir=/usr/lib/gcc/x86_64-linux-gnu/9 -fPIC -std=c99 -Wsign-conversion -Wall -Wextra -pedantic -DUTF8PROC_EXPORTS  -c -o utf8proc.o utf8proc.c
clang-16: error: '/usr/lib/gcc/x86_64-linux-gnu/9' does not contain a GCC installation
```

On the previous commit I saw
```
Found candidate GCC installation: /usr/local/bin/../lib/gcc/x86_64-linux-gnu/9.1.0//../../../../lib/gcc/x86_64-linux-gnu/9.1.0
Selected GCC installation: /usr/local/bin/../lib/gcc/x86_64-linux-gnu/9.1.0//../../../../lib/gcc/x86_64-linux-gnu/9.1.0
```
So I presume the specified path was wrong.

Zentrik:
Thanks, it seems to be working much better now. I just need to figure out why there's a hash mismatch, I had the same locally but I just refreshed it.

pipelines/main/platforms/build_linux.pgo_lto.arches
```diff
@@ -0,0 +1,8 @@
+# ROOTFS_IMAGE_NAME    TRIPLET                        ARCH           ARCH_ROOTFS    USE_JULIA_PGO_LTO   MAKE_FLAGS                                                                              TIMEOUT       ROOTFS_TAG    ROOTFS_HASH
+llvm_passes          x86_64-linux-gnu               x86_64         x86_64         PGO_LTO    JL_CFLAGS=-Werror -Wno-error=unused-function,JL_CXXFLAGS=-Werror -Wno-error=unused-function                                                   .             v6.00         0badf5d8794e21a03ac2ff2d46c1ab538ac02571
```
giordano, line 2: It may be better to address the warning rather than ignoring it.

Zentrik: Was a bug in clang, workaround here: https://github.com/JuliaLang/julia/pull/53897.

Zentrik:
This looks good to me, testing was 15% faster. If we want `-Werror` I think we just need to merge or otherwise fix https://github.com/JuliaLang/julia/pull/53897 and bump libwhich to https://github.com/vtjnash/libwhich/commit/613b3f00078917d91b03b5734c6cbb183f20e72d.

pipelines/main/launch_unsigned_jobs.yml
```diff
@@ -30,6 +30,11 @@ steps:
               bash .buildkite/utilities/arches_pipeline_upload.sh \
               .buildkite/pipelines/main/platforms/build_linux.arches \
               .buildkite/pipelines/main/platforms/build_linux.yml
+          GROUP="Build" \
```
Zentrik, line 33:
```suggestion
	  # Launch Linux PGO LTO build jobs
          GROUP="Build" \
```

pipelines/main/launch_unsigned_jobs.yml
```diff
@@ -30,6 +30,12 @@ steps:
               bash .buildkite/utilities/arches_pipeline_upload.sh \
               .buildkite/pipelines/main/platforms/build_linux.arches \
               .buildkite/pipelines/main/platforms/build_linux.yml
+	  # Launch Linux PGO LTO build jobs
```
Zentrik, line 33:
```suggestion
          # Launch Linux PGO LTO build jobs
```

giordano:
I discussed this with @Zentrik a few days ago: I believe it'd be great to actually test the PGO+LTO pipeline, but I'm a bit concerned with the extra long job this introduces.  One option could be to move the job as a scheduled one, with the counter effect few eyes will look at it and the risk this will bitrot at some point.  What do people think?

Zentrik:
In terms of speeding up the build job we could build the instrumented LLVM on yggy. Further, we could only profile periodically, e.g. every LLVM upgrade, allowing us to skip the slow profiling step and to distribute the optimised LLVM from yggy.

PR #346, opened 2024-03-27
Upload ASAN build artifacts

### LINKED ISSUE #169: Upload ASAN artifacts
Building Julia with ASAN is non-trivial and takes a significant amount of time. Since we build it for CI already, it would be nice if we go one step further and package the build + upload as an artifact. This can be helpful when debugging failures, e.g., looking at https://github.com/JuliaLang/julia/pull/45608 the PkgEval logs showed some possibly memory corruption-related segfaults, for which it would have been helpful to download the ASAN build and try a `Pkg.test` locally.

### LINKED ISSUE #331: Upload build artifacts for `asan` and `tsan`
cc @vchuravy @Keno (happy new years!)

Referencing comment from a year ago – https://github.com/JuliaLang/julia/issues/48031#issuecomment-1379236565

I was wondering if tsan and asan builds of Julia could have build artifacts available for download? I've struggled with building these in the past. It would be very useful for users like myself who just want to quickly debug a data race without having to configure a tsan build.

Zentrik:
Closes #169, partially deals with #331.
Hopefully this works

### DIFF
--- a/pipelines/main/misc/sanitizers/asan.yml
+++ b/pipelines/main/misc/sanitizers/asan.yml
@@ -32,7 +32,7 @@ steps:
         source .buildkite/utilities/build_envs.sh
 
         cd ./tmp/test-asan/asan
-        UPLOAD_FILENAME = "${JULIA_BINARYDIST_FILENAME}-asan"
+        UPLOAD_FILENAME +=-asan
 
         # Rename the build artifact in case we want to name it differently, as is the case on `musl`.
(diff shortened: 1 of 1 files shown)


Zentrik:
Note this only uploads to buildkite not S3.

Zentrik:
Looks like it segfaulted on `Pkg.instatiate()` when building the docs.

Zentrik:
Looks like `binary-dist` doesn't work for ASAN, e.g. we don't copy over `libclang_rt.asan-x86_64.so`.

PR #348, opened 2024-04-08
Remove trailing slash in "Official https://julialang.org/ release"

LilithHafner:
Pairs with https://github.com/JuliaLang/julia/pull/53978


LilithHafner:
We should only make this change in Julia 1.12+ because previous versions expect `Base.TAGGED_RELEASE_BANNER == "Official https://julialang.org/ release"`

DilumAluthge:
> We should only make this change in Julia 1.12+

Probably the simplest way is to just have an `if` in this Bash script, and do something like sed or grep the VERSION file.

@gbaraldi What do you think?

The more complicated way would be to make some `release-` branches on this repo (the `julia-buildkite` repo). That's a more complex approach that brings more overhead.

LilithHafner:
`grep -P` is not supported. Refactoring.

LilithHafner:
```
Make Options:
 -> -j4
 -> VERBOSE=1
 -> TAGGED_RELEASE_BANNER=Official https://julialang.org release
```

Seems to be working. Do we test building old versions of Julia in CI here?

DilumAluthge:
> Do we test building old versions of Julia in CI here?

Currently, in this repo, PRs against the `main` branch of this repo will only test building the `master` branch of Julia.

It is possible to do a test against a different branch of Julia. I don't remember the steps off the top of my head, but @gbaraldi might remember.

utilities/build_julia.sh
```diff
@@ -33,7 +33,12 @@ MFLAGS+=( "-j${JULIA_CPU_THREADS}")
 
 # Add a few default flags to our make flags:
 MFLAGS+=( "VERBOSE=1" )
-MFLAGS+=( "TAGGED_RELEASE_BANNER=Official https://julialang.org/ release" )
+if grep '1\.\(1[2-9]\|[2-9][0-9]\|[0-9][0-9][0-9][0-9]*\)\.' VERSION >/dev/null; then # if version >= 1.12
```
staticfloat, line 36: We can make things a little less `awk`ward:

```suggestion
# NOTE: this breaks if you build v1.9 or lower with this
if awk "BEGIN {exit !($d1 >= 1.12)}"; then
```

We're not going to build v1.9 with this code, so the numeric comparison is fine.

LilithHafner: Not a fan of this change, but don't really care.

LilithHafner:
```suggestion
# Taken from https://stackoverflow.com/a/4024263
verlte() {
    printf '%s\n' "$1" "$2" | sort -C -V
}
verlt() {
    ! verlte "$2" "$1"
}
if verlt "1.12" "$(cat VERSION)"; then
```

Issue #336, opened 2024-02-04
Documenter making builds fail on julia-buildkite repo CI


IanButterworth:
https://github.com/JuliaCI/julia-buildkite/pull/332#issuecomment-1917950993

Making this an issue so we can track it. 
cc. @mortenpi

mortenpi:
I would put the "blame" here on the CI scripts though. It looks like the [code is getting checked out in a weird way](https://github.com/JuliaCI/julia-buildkite/pull/332#issuecomment-1918160653), and so it breaks `doc/make.jl`, which assume a standard checkout. Of course, we might be able to adapt `doc/make.jl` to make this work somehow, but I don't really have a sense what the differences are between a standard Julia checkout and what the CI does.

PR #352, opened 2024-05-09
Codesign windows tarballs as well

staticfloat:
Because the codesigning of files is done by the Inno installer build, we need to repackage the tarball after the build is done, just like we do for macOS.


staticfloat:
This will solve https://github.com/JuliaLang/julia/issues/54365

DilumAluthge (approved):

DilumAluthge:
cc: @davidanthoff

DilumAluthge:
@staticfloat @mortenpi @gbaraldi We need to figure out a way to fix the Documenter bug in the CI on this repo. Otherwise, CI is always red, which increases the risk of us merging a true failure.

staticfloat:
Yes, we actually need to solve it for this PR because I'm trying to test something related to Windows.  @mortenpi is there some kind of override I need to pass Documenter to trick it into thinking that the repository URL is actually JuliaLang/julia?

mortenpi:
X-ref https://github.com/JuliaCI/julia-buildkite/issues/336#issuecomment-1925995483. But I hope that https://github.com/JuliaLang/julia/pull/54445 will fix it. Can someone test that? It fixes the case when you call `make docs` on an unpacked tarball.

DilumAluthge:
CI is all green now (except for the scheduled pipeline, which has been red for a long time now).

DilumAluthge:
Thanks @mortenpi for the fix!

staticfloat:
@davidanthoff can you verify that the tarballs uploaded by the upload jobs here are sufficiently signed?  The links are:

* https://buildkite.com/organizations/julialang/pipelines/julia-buildkite/builds/1511/jobs/018f7325-f9ca-4606-822a-d994b8e521b3/artifacts/018f737c-c615-492c-a6f6-4bded473cf41
* https://buildkite.com/organizations/julialang/pipelines/julia-buildkite/builds/1511/jobs/018f7325-f6e3-40d5-9a03-d051a152f736/artifacts/018f7379-4084-4396-a295-bb18bed1a574

davidanthoff:
Both archives now have `bin/julia.exe` signed, so we are definitely back to where we were a while ago!

But none of the dlls are signed, and I'm wondering whether some of the problems in https://github.com/JuliaLang/julia/issues/54366 might go away if we just sign everything, including all the dlls in `bin` as well as the sysimage, i.e. `lib/julia/sys.dll`. The theory there being that Windows Defender is slowing things down somehow, and maybe it would take a less involved path for files that have a valid signature on it.

Also the files in `libexec/julia` are also not signed, but that is probably ok?

Oh, and I just realized that there are also lots of shared libraries in `share/julia/compiled/v1.12/`... If this theory about Windows Defender is correct, then we should probably sign all of those as well...

davidanthoff:
No, I don't think dll files were ever signed anywhere, I just think ideally they should be :)

DilumAluthge:
Can we go ahead and merge this PR (since it does improve the situation a little bit), and then open an issue for the follow-up task (signing the DLLs inside both the `.tar.gz` and `.zip`)?

DilumAluthge (approved):
CI is fixed now.

DilumAluthge:
I've opened this issue for the follow-up task (signing the DLLs): #353

DilumAluthge:
> Both archives now have `bin/julia.exe` signed, so we are definitely back to where we were a while ago!

@davidanthoff What command did you run to verify the codesigning?

I was thinking it would be nice for us to run that command in our CI script, so that we can detect if this regresses in the future.

PR #354, opened 2024-05-17
only use apple-m1 target for macos

IanButterworth:
Came up as part of https://github.com/JuliaLang/julia/pull/54471

Master
MacOS aarch64: 5m55s 214 MB artifact

This PR
MacOS aarch64: 5m5s 197 MB artifact

### DIFF
--- a/utilities/build_envs.sh
+++ b/utilities/build_envs.sh
@@ -54,21 +54,30 @@ case "${ARCH?}" in
         )
         ;;
     aarch64)
-        JULIA_CPU_TARGETS+=(
-            # Absolute base aarch64 feature set
-            "generic"
-            # Cortex A57, Example: NVIDIA Jetson TX1, Jetson Nano
-            "cortex-a57"
-            # Cavium ThunderX2T99, a common server architecture
-            "thunderx2t99"
-            # NVidia Carmel, e.g. Jetson AGX Xavier; serves as a baseline for later architectures
-            "carmel,clone_all"
-            # Apple M1
-            "apple-m1,base(3)"
-            # Vector-length-agnostic common denominator between Neoverse V1 and V2, recent Arm server architectures
-            "neoverse-512tvb,base(3)"
-        )
-        ;;
+        case "${OS?}" in
+            macos)
+                JULIA_CPU_TARGETS+=(
+                    # Absolute base aarch64 feature set
+                    "generic"
+                    # Apple M1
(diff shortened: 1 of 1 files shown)


utilities/build_envs.sh
```diff
@@ -18,7 +18,7 @@ if buildkite-agent meta-data exists BUILDKITE_JULIA_BRANCH; then
 fi
 
 # Determine JULIA_CPU_TARGETS for different architectures
-JUlIA_CPU_TARGETS=()
+JULIA_CPU_TARGETS=()
```
IanButterworth, line 21: How was this not failing??

utilities/build_envs.sh
```diff
@@ -54,20 +54,30 @@ case "${ARCH?}" in
...
+                JULIA_CPU_TARGETS+=(
+                    # Absolute base aarch64 feature set
+                    "generic"
+                    # Cortex A57, Example: NVIDIA Jetson TX1, Jetson Nano
+                    "cortex-a57"
+                    # Cavium ThunderX2T99, a common server architecture
+                    "thunderx2t99"
+                    # NVidia Carmel, e.g. Jetson AGX Xavier; serves as a baseline for later architectures
+                    "carmel,clone_all"
+                    # Vector-length-agnostic common denominator between Neoverse V1 and V2, recent Arm server architectures
```
giordano, line 76: Why not apple-m1?

IanButterworth: ah-sorry (asahi), I forgot about using other OSs on macs

utilities/build_envs.sh
giordano: Because `bash` doesn't have problems with `+=` for an undefined variable.  I don't know if there's a mode to turn that into an error, `set -u` doesn't seem to help here.

utilities/build_envs.sh
giordano: I haven't checked carefully, but it may also be usable by Neoverse (since the `neoverse-512tvb` target isn't doing anything right now)

IanButterworth:
Does this look alright @staticfloat @gbaraldi ?

staticfloat (approved):

DilumAluthge:
Looks like CI was failing in this PR, because the tree ashes needed to be re-signed.

DilumAluthge:
I've reverted this PR in #355, to get Julia master green again.

PR #357, opened 2024-06-09
reland "only use apple-m1 target for macos (#354)"

IanButterworth:
#354 wasn't signed before merge so was reverted


IanButterworth:
Can I get a hand signing this please @DilumAluthge @staticfloat 

(Or explain what I need to do)

IanButterworth:
#354 was approved and this is passing CI, except for failures that are on master too, so I'll go ahead

PR #363, opened 2024-06-19
Upload on PR builds as well

staticfloat:
This is a follow-up to c8b470989226b3a8f8d4874268ccbee616da2ebf, I forgot that we elide the upload jobs completely when not on `master` or `release-*`.  This remediates the situation, uploading PR builds as well.


DilumAluthge:
So, with this PR, we'll start uploading all PR builds to S3, right?

Our S3 usage is going to increase by a not-insignificant amount. Is that a concern?

staticfloat:
> Our S3 usage is going to increase by a not-insignificant amount. Is that a concern?

It will probably increase our monthly bill by $200-300.  It's not too bad.

DilumAluthge:
Should we also have an S3 lifecycle rule to automatically delete PR builds after `N` days?

staticfloat:
All builds uploaded to the nightlies bucket are deleted after 60 days.

DilumAluthge (dismissed):

DilumAluthge (dismissed):

DilumAluthge (approved):

PR #374, opened 2024-07-31
Codesign all executable objects in windows build

staticfloat:
Inno setup does its own codesigning which is great and all, but we want code signatures on our `.exe` and `.dll` files in our `.zip` and `.tar.gz` distributions as well.


utilities/upload_julia.sh
```diff
@@ -17,7 +17,7 @@ UPLOAD_EXTENSIONS=( "tar.gz" )
 
 # Only codesign if we are not on a pull request build.
 # Pull request builds only upload unsigned tarballs.
-if [[ "${BUILDKITE_PULL_REQUEST}" == "false" ]]; then
+#if [[ "${BUILDKITE_PULL_REQUEST}" == "false" ]]; then
```
KristofferC, line 20: Was this intended? Should it be removed then?

DilumAluthge: That change is in a commit marked "DROP ME", so I think it's just a temp change while debugging this PR.

KristofferC: Ah, heh, sorry for the noise.

utilities/windows/codesign.sh
```diff
@@ -72,27 +72,52 @@ SERVERS=(
 NUM_RETRIES=3
 
 function do_codesign() {
+    set -x
     for retry in $(seq 1 ${NUM_RETRIES}); do
         for SERVER in ${SERVERS[@]}; do
             # Note that we're using SHA1 signing here, because that's what our certificate supports.
             # In the future, we may be able to upgrade to SHA256.
-            if signtool sign /debug /fd SHA1 /f "${CERT_PATH}" /p "${CERT_PASSWORD}" /t "${SERVER}" "$1"; then
+            if signtool sign /debug /fd certHash /f "${CERT_PATH}" /p "${CERT_PASSWORD}" /t "${SERVER}" "$1"; then
```
Keno, line 80: Try windows filepaths for $1? Support for unix path separators is inconsistent in windows utilities

davidanthoff:
We got the trusted signing account working now, the GitHub trusted signing action is successfully signing the Juliaup stuff.

So I think the best strategy here is to use https://www.powershellgallery.com/packages/TrustedSigning/0.4.1 for signing and just completely ditch `signtool`. My sense is that is the easiest path forward.

One _can_ also use `signtool` with trusted signing, but the instructions https://learn.microsoft.com/en-us/azure/trusted-signing/how-to-signing-integrations seem very involved and much more complicated than the PowerShell module.

utilities/windows/codesign.sh
vtjnash: We look exactly like https://gist.github.com/brimston3/3839a1e112fd207152569bfc99e0d7ef

Is is possibly better to just specify `SHA1` explicitly instead of making it guess that from `/f`?

staticfloat: I had it as `SHA1` before, that didn't work either.  The thing that really confuses me is that it works when invoked from `inno`, but not here.

utilities/windows/codesign.sh
```diff
@@ -77,7 +77,7 @@ function do_codesign() {
         for SERVER in ${SERVERS[@]}; do
             # Note that we're using SHA1 signing here, because that's what our certificate supports.
             # In the future, we may be able to upgrade to SHA256.
-            if signtool sign /debug /fd certHash /f "${CERT_PATH}" /p "${CERT_PASSWORD}" /t "${SERVER}" "$1"; then
+            if MSYS2_ARG_CONV_EXCL='*' signtool sign /debug /fd certHash /f "${CERT_PATH}" /p "${CERT_PASSWORD}" /t "${SERVER}" "$1"; then
```
Keno, line 80: That's quite the trap ...

IanButterworth:
I think this has caused re-precompilation because the `.ji` file checksums aren't updated after the codesigning like we do on MacOS
```
┌ Debug: Rejecting cache file C:\hostedtoolcache\windows\julia\nightly\x64\share\julia\compiled\v1.12\REPL\u0gqU_Miz3B.ji because C:\hostedtoolcache\windows\julia\nightly\x64\share\julia\compiled\v1.12\REPL\u0gqU_Miz3B.dll has an invalid checksum
└ @ Base loading.jl:3868
```

PR #376, opened 2024-08-15
update stdlib pkgimage checksums for windows

staticfloat:


utilities/upload_julia.sh
```diff
@@ -91,6 +91,9 @@ if [[ "${BUILDKITE_PULL_REQUEST}" == "false" ]]; then
         echo "--- [windows] Codesign everything in the install directory"
         "${codesign_script}" --certificate="${certificate}" "${JULIA_INSTALL_DIR}"
 
+        echo "--- [mac] Update checksums for stdlib cachefiles"
```
IanButterworth, line 94:
```suggestion
        echo "--- [windows] Update checksums for stdlib cachefiles"
```

staticfloat:
@IanButterworth this looks to me like it's working, can you confirm?

utilities/update_stdlib_pkgimage_checksums.jl
```diff
@@ -0,0 +1,99 @@
+using Libdl
+import Base: _sizeof_uv_fs, uv_error
+
+# This should be in Julia base, I feel
```
vchuravy, line 4: Yeah this should be in contrib, so that we update it there when the header format changes

IanButterworth:
LGTM from the logs here 👍

utilities/update_stdlib_pkgimage_checksums.jl
IanButterworth: Sounds good. Perhaps could be a follow-on PR?

utilities/update_stdlib_pkgimage_checksums.jl
```diff
@@ -0,0 +1,99 @@
...
+
+    # respect mtime order of caches, so that if one cache depends on another, we end up
+    sort!(pkgimg_files, by=mtime)
+    for pkgimg_file in pkgimg_files
+        ji_file = string(splitext(pkgimg_file)[1], ".ji")
+        preserve_mtimes([ji_file, pkgimg_file]) do
+            update_cache_pkgimg_checksum!(ji_file, pkgimg_file)
+        end
+    end
+end
```
IanButterworth, line 99:
```suggestion
end

# check stdlibs caches are valid
Base.isprecompiled(Base.PkgId(Base.UUID("44cfe95a-1eb2-52ea-b672-e2afdf69b78f"), "Pkg")) || error()
```

IanButterworth:
I added a check to the script, Once this is green we can drop [`285ec94` (#376)](https://github.com/JuliaCI/julia-buildkite/pull/376/commits/285ec9451db263cceec7f57e46c05f4d108ca01d) and merge, IMO

utilities/update_stdlib_pkgimage_checksums.jl
```diff
@@ -0,0 +1,102 @@
...
+
+# check stdlibs caches are valid
+Base.isprecompiled(Base.PkgId(Base.UUID("44cfe95a-1eb2-52ea-b672-e2afdf69b78f"), "Pkg")) || error()
```
KristofferC, line 102: This check fails on the 1.10 build for some reason.

IanButterworth: Change it to delimitedfiles? I think that's the only excised stdlib on 1.10?

PR #380, opened 2024-08-20
try fix build on 1.10 by checking DelimitedFiles (which is not in the sysimage) for being precompiled

KristofferC:


KristofferC:
Progress, it seems to complain about a new commit now.

PR #383, opened 2024-08-26
Set `JULIA_DEBUG=all` when updating stdlib checksums

staticfloat:


staticfloat:
Looks like we [do get some interesting info here](https://buildkite.com/julialang/julia-buildkite/builds/1755#01918ff9-01b1-4b1f-9e0e-c617ccaa5dd1/781-888) so I'm going to merge this, and hopefully this will help us debug why we're having trouble with v1.10

PR #386, opened 2024-09-14
Add Worldwide Developer Relations certificate to keychain as well

gbaraldi:


giordano:
Need to update the devdocs?

gbaraldi:
I don't think so? This isn't something that changes

PR #372, opened 2024-07-30
Test Revise on buildkite

KristofferC:
It's been discussed that Revise should also be tested on CI because the effect of the productivity of Julia itself significantly goes down when Revise is not working.

Also:

image

Someone feel free to take over this PR.

PR #393, opened 2024-10-04
Mark the `test-revise` job as required

DilumAluthge:
Follow-up to #372.

This PR makes the `test-revise` job required. Thus, if the `test-revise` job fails, the overall build will fail.


DilumAluthge:
Closing in favor of #394, which is more comprehensive.

PR #395, opened 2024-10-09
use v2 of staticfloat/sandbox that has newer manifest that works with 1.11+

KristofferC:


KristofferC:
Sandbox environment instantiates successfully here.

PR #397, opened 2024-10-11
Bump linux timeouts

Keno:
We're seeing some intermittent timeouts on linux master CI. However, successful runs are pretty close to the limit (our tests have started to grow over time), so try bumping the timeout. Of course it's possible we do actually have some sort of hang.

### DIFF
--- a/pipelines/main/platforms/test_linux.arches
+++ b/pipelines/main/platforms/test_linux.arches
@@ -6,7 +6,7 @@ tester_linux           x86_64-linux-gnuassert     x86_64         x86_64         
 # These special lines allow us to embed default values for the columns above.
 # Any column without a default mapping here will simply substitute a `.` to the empty string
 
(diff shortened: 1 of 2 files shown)


DilumAluthge:
It's been a bit since I looked at these files. @staticfloat Are these the right places to increase the timeouts? Do we also need to update any other locations?

staticfloat:
Yes, that will update the normal Linux builder without changing any of the others in that arch file.

staticfloat (approved):

DilumAluthge:
@staticfloat 

> ERROR: LoadError: IOError: realpath("/cache/sandbox_persistence"): no such file or directory (ENOENT)

Keno:
That was a transient issue with the demeter6 workers. I believe @fredrikekre fixed those.

DilumAluthge:
Cool - I've triggered retries of the failed jobs, so hopefully they'll pass now.

PR #405, opened 2024-11-14
Disable PowerPC CI

Keno:
And though we reached the summit, we did not discover salvation, only misery....

Summit is being turned off tomorrow. As far as we know, that was the last PowerPC platform that anybody was seriously using Julia on. CI has been broken on master recently and nobody seems particularly inclined to fix it, so turn off CI and we'll drop down PowerPC to Tier 4.


DilumAluthge:
How about we disable it for Julia 1.11 and up, but keep it for Julia 1.10.x?

giordano:
I don't think we have separate pipelines for different versions, not yet at least. And it should work on v1.11

DilumAluthge:
> I don't think we have separate pipelines for different versions, not yet at least. And it should work on v1.11

Alright, #407 keeps PowerPC for both 1.10 and 1.11. And instead of maintaining completely separate Buildkite pipelines, we just use a conditional.

DilumAluthge:
Closing in favor of #407.

PR #407, opened 2024-11-18
Disable PowerPC CI for Julia 1.12+ (but retain it for Julia prior to 1.12)

DilumAluthge:
Alternative to #405.

If Julia is 1.12+, we don't run PowerPC.

For Julia prior to 1.12, we do run PowerPC.

PR #400, opened 2024-11-02
install diff-filter-buildkite-plugin to only build when relevant files changed

fatteneder:
This adds a filter that skips CI if a PR only changed things like `README.md, HISTROY.md` etc.

This is a first attempt towards fixing #243.
Still lacks a few desired features like (list taken from the recent Slack thread, @DilumAluthge)
- manually ignoring the filter for pre-releases and tags,
- ignore the filter when merging into `master`.

PR #412, opened 2024-12-02
`launch_powerpc.jl`: Forward env vars to upload script

fatteneder:
Found this while rebasing #400.

This might address the build failures mentioned here: https://github.com/JuliaCI/julia-buildkite/pull/407#issuecomment-2481823956


DilumAluthge:
Hmmm... I think Julia should already be forwarding the environment in this case...

fatteneder:
I tried my MWE again and realized that I forgot to export the relevant variable, my bad.

PR #413, opened 2024-12-10
do not run Pkg tests on rr

KristofferC:
rr has problems with Pkg tests


DilumAluthge:
Yeah, we likely need to do this for now, to make CI green again.

As Kristoffer mentioned in a different thread (JuliaLang/julia#56781), the Pkg tests pass in the regular job (no rr), and fail under rr, so presumably this is an rr bug (not a Pkg bug).

@Keno @vtjnash It might be worth seeing if that rr bug can be fixed? We do have rr traces from the failed jobs - see e.g. the "Artifacts" tab of the `test x86_64-linux-gnuassertrr-net` job here: https://buildkite.com/julialang/julia-master/builds/42912#0193b106-6c7d-453b-84d9-3f55a4a6d6c2

DilumAluthge (approved):

PR #414, opened 2024-12-10
Prevent some "launch jobs" jobs from running on pibots

DilumAluthge:


DilumAluthge:
cc: @topolarity

fredrikekre (approved):

topolarity:
Should we do the same for "Test" and "Check"?

topolarity:
> Should we do the same for "Test" and "Check"?

I guess that's not really necessary, since the other jobs don't use Julia

DilumAluthge:
Yeah, it's the need to install Julia that causes problems on the pibots.

PR #415, opened 2024-12-30
Shorten make paths in logs, with key at start of log

IanButterworth:
Because they're so long that it makes logs hard to read.

Before
Screenshot 2024-12-29 at 9 40 13 PM

This PR
Screenshot 2024-12-29 at 9 41 58 PM


DilumAluthge (approved):
SGTM once `[julia]` is changed to `[buildroot]`.

DilumAluthge (approved):

PR #419, opened 2025-01-19
Adding CI for Julia+MMTk

udesou:
This PR adds a build and test CI job for building and testing Julia with MMTk as the default GC.


inkydragon:
The build system does not recognize the new triplet.
```
$ trap 'kill -- $$' INT TERM QUIT; bash .buildkite/utilities/build_julia.sh
Unknown triplet OS 'x86_64-linux-gnummtk'
🚨 Error: The command exited with status 1
```
https://buildkite.com/julialang/julia-buildkite/builds/1948#019480f7-738c-4fce-9ed2-7ae0cc5c5f22/630-632

Need to add a new case here:
https://github.com/JuliaCI/julia-buildkite/blob/6c905bfb9349d45ea7c512837186208a832ef8b5/utilities/extract_triplet.sh#L23-L29

udesou:
@inkydragon Any chance you can sign the files so we can get all tests passing (and merge the PR)? Thank you!!

inkydragon:
I don't have access to sign for the pipeline.
ping @DilumAluthge

gbaraldi (approved):

DilumAluthge:
I can re-sign the pipelines sometime this week, unless someone beats me to it.

staticfloat (approved):

d-netto:
Failures in `Test` seem unrelated? It's failing on Windows, which is not tested my MMTk.

DilumAluthge:
I've retriggered the failed build.

IanButterworth:
Seems like the new job is using the same name as another for its artifacts, breaking the upload job.

udesou:
I've tried to fix it. Let's see if that works.

udesou:
Looks like the pipeline will need re-signing though.

d-netto:
Bump?

DilumAluthge:
Sometimes we have to remove and re-add the `automerge` label in order for Kodiak to pick up the latest CI commit statuses.

Keno:
Can we allow-fail this one while mmtk is still experimental? In particular, the problem I have is that changes to the scanning code now need a corresponding change in mmtk-julia, which needs a whole jll update cycle. I think we at the very least want to have that code in the Julia repo before we make it a mandatory CI check.

giordano:
Also, presumably this should run only on `master` (and maybe the 1.12 branch)?

DilumAluthge:
Allow-fail seems reasonable. Which jobs should be allowed to fail? Just the tests, or both the build and the test?

Running on only 1.12+ also seems reasonable.

PR #420, opened 2025-01-24
Upload after build without waiting for tests

ericphanson:
I propose this change because currently if tests fail, the build does not upload. I think some of the point of using nightly/pr-builds is to test things manually, and potentially to debug failing tests and issues like that. Therefore, I don't think uploads should be gated on tests passing.

### DIFF
--- a/pipelines/main/platforms/upload_freebsd.yml
+++ b/pipelines/main/platforms/upload_freebsd.yml
@@ -6,8 +6,6 @@ steps:
         depends_on:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
-          # Wait for the tester to finish
-          - "test_${TRIPLET?}"
         # Prevent multiple pipelines from uploading to S3 simultaneously
         # It is okay for two different triplets to upload simultaneously
         concurrency: 1
--- a/pipelines/main/platforms/upload_linux.yml
+++ b/pipelines/main/platforms/upload_linux.yml
@@ -6,8 +6,6 @@ steps:
         depends_on:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
-          # Wait for the tester to finish
-          - "test_${TRIPLET?}"
         # Prevent multiple pipelines from uploading to S3 simultaneously
         # It is okay for two different triplets to upload simultaneously
         concurrency: 1
--- a/pipelines/main/platforms/upload_macos.yml
+++ b/pipelines/main/platforms/upload_macos.yml
@@ -6,8 +6,6 @@ steps:
         depends_on:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
-          # Wait for the tester to finish
-          - "test_${TRIPLET?}"
         # Prevent multiple pipelines from uploading to S3 simultaneously
         # It is okay for two different triplets to upload simultaneously
         concurrency: 1
--- a/pipelines/main/platforms/upload_windows.yml
+++ b/pipelines/main/platforms/upload_windows.yml
@@ -6,8 +6,6 @@ steps:
         depends_on:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
-          # Wait for the tester to finish
-          - "test_${TRIPLET?}"
         # Prevent multiple pipelines from uploading to S3 simultaneously
         # It is okay for two different triplets to upload simultaneously
         concurrency: 1


DilumAluthge:
We do want to gate it for releases. Can we make the behavior different for nightlies vs releases?

ericphanson:
ah I see. I am having trouble figuring out how to do that. I think we want the conditional `(build.branch =~ /^release-/) || (build.tag =~ /^v/)` or similar somewhere, but I can't figure out how to get a conditional dependency, or a conditional allowed-failure... 🤔

ericphanson:
I gave it a try, but I don't know if it will work -- in particular, I don't know if ENV interpolation works like this for `allow_failures`

staticfloat:
If the concern is that you want to be able to download a build of Julia that is failing tests so you can inspect it, you can download it from the buildkite interface in the Artifacts tab.  I don't think there's much use uploading these to S3, is there?

IanButterworth:
Sometimes there are PRs that don't yet have all the tests figured out, but would be good for people to try out. The `--trim` PR was one of those.

giordano:
> I don't think there's much use uploading these to S3, is there?

My guess is that this would be useful for juliaup?  I'm not sure where it downloads the tarball from, but if it's S3 that'd be a discriminant here

ericphanson:
Yeah, my gripe here is I wanted to use the Pkg apps feature before https://github.com/JuliaLang/julia/pull/57187 was merged, but I couldn't use the juliaup PR feature for it since the macos CI had failed. It was confusing initially as well bc juliaup reports it can't be installed, but I saw the PR is there and didn't understand that uploads were gated on CI passing.

staticfloat:
Fair enough, the juliaup feature makes this much more useful.

DilumAluthge:
I don't think this is ready to merge yet? We need to make it so that release branches and tags DO gate on tests, but nightlies and PRs do NOT gate on tests.

It looks like this PR attempts to implement that, but something about the implementation is causing CI to fail.

DilumAluthge:
Also needs treehash re-signing.

DilumAluthge:
> fatal: pipeline interpolation of "upload_linux.yml" failed: interpolating remaining fields: $ALLOW_TEST_FAILURES: not set

ericphanson:
looks like I missed a codepath for the allowed-to-fail upload jobs, should be covered now

DilumAluthge:
Before we re-sign the pipelines, could you add comments that explain the purpose of the logic (i.e. that we want to gate on the tests in release branches and tags, but we don't want to gate on the tests in master/nightlies/prs).

ericphanson:
added

IanButterworth:
@ericphanson @DilumAluthge lets merge this, if you agree it's still in good shape?

pipelines/main/launch_upload_jobs.yml
```diff
@@ -23,6 +23,10 @@ steps:
           export BUILDKITE_PLUGIN_CRYPTIC_BASE64_SIGNED_JOB_ID_SECRET
           export ALLOW_FAIL="false"
 
+          # allow test-failures for non-release/tags so that nightly & PR builds
+          # can be uploaded to s3 even if tests are not passing
+          # (allowing for usage with juliaup)
+          export ALLOW_TEST_FAILURES=$(([[ "${BUILDKITE_BRANCH?}" == release-* ]] || [[ "${BUILDKITE_TAG:-}" == v* ]]) && echo "false" || echo "true")
```
DilumAluthge, line 29: I haven't actually verified myself that this pattern matching is correct, so ideally someone would just do some quick tests in Bash locally before we merge.

Or alternatively, we can have a quick GitHub Actions CI job to check miscellaneous things such as this. E.g. just a quick Bash-based CI job that puts in various values of `BUILDKITE_BRANCH` and `BUILDKITE_TAG`, and tests that the output `ALLOW_TEST_FAILURES` is correct.

The latter is probably easier, because we can do it in CI instead of relying on people to manually test things locally.

pipelines/main/launch_upload_jobs.yml
```diff
@@ -84,6 +88,11 @@ steps:
           export BUILDKITE_PLUGIN_CRYPTIC_BASE64_SIGNED_JOB_ID_SECRET
 
           export ALLOW_FAIL="true"
+          
+          # allow test-failures for non-release/tags so that nightly & PR builds
+          # can be uploaded to s3 even if tests are not passing
+          # (allowing for usage with juliaup)
+          export ALLOW_TEST_FAILURES=$(([[ "${BUILDKITE_BRANCH?}" == release-* ]] || [[ "${BUILDKITE_TAG:-}" == v* ]]) && echo "false" || echo "true")
```
DilumAluthge, line 95: Same as above - would be nice to quickly test in CI (in a small GitHub Actions CI job).

pipelines/main/launch_upload_jobs.yml
DilumAluthge, line 29: Also, can we rename this something else? In Buildkite, "allow fail" (or "soft fail") means something different - it means failures show up as a red warning sign (not a red X), and failures don't block `Test` from being green.

Here, we have a different meaning, so can we make the variable name more unambiguous?

```suggestion
          export UPLOAD_EVEN_IF_TEST_FAILURES=$(([[ "${BUILDKITE_BRANCH?}" == release-* ]] || [[ "${BUILDKITE_TAG:-}" == v* ]]) && echo "false" || echo "true")
```

pipelines/main/platforms/upload_freebsd.yml
```diff
@@ -7,7 +7,8 @@ steps:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
           # Wait for the tester to finish
-          - "test_${TRIPLET?}"
+          - step: "test_${TRIPLET?}"
+            allow_failure: "${ALLOW_TEST_FAILURES?}"
```
DilumAluthge, line 11:
```suggestion
            allow_failure: "${UPLOAD_EVEN_IF_TEST_FAILURES?}"
```

pipelines/main/platforms/upload_linux.yml
```diff
@@ -7,7 +7,8 @@ steps:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
           # Wait for the tester to finish
-          - "test_${TRIPLET?}"
+          - step: "test_${TRIPLET?}"
+            allow_failure: "${ALLOW_TEST_FAILURES?}"
```
DilumAluthge, line 11:
```suggestion
            allow_failure: "${UPLOAD_EVEN_IF_TEST_FAILURES?}"
```

pipelines/main/platforms/upload_macos.yml
```diff
@@ -7,7 +7,8 @@ steps:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
           # Wait for the tester to finish
-          - "test_${TRIPLET?}"
+          - step: "test_${TRIPLET?}"
+            allow_failure: "${ALLOW_TEST_FAILURES?}"
```
DilumAluthge, line 11:
```suggestion
            allow_failure: "${UPLOAD_EVEN_IF_TEST_FAILURES?}"
```

pipelines/main/platforms/upload_windows.yml
```diff
@@ -7,7 +7,8 @@ steps:
           # Wait for the builder to finish
           - "build_${TRIPLET?}"
           # Wait for the tester to finish
-          - "test_${TRIPLET?}"
+          - step: "test_${TRIPLET?}"
+            allow_failure: "${ALLOW_TEST_FAILURES?}"
```
DilumAluthge, line 11:
```suggestion
            allow_failure: "${UPLOAD_EVEN_IF_TEST_FAILURES?}"
```

DilumAluthge (commented):
Two requests:

1. Rename variable name to make it more clear that this is specifically just affecting whether or not we upload - it doesn't affect anything else.
2. Ideally, I'd like to have a small GitHub Actions CI job that runs a Bash script to just test that our Bash syntax for pattern matching is correct, and that it gives the expected output when the branch is `release-something` or the tag is `vSOMETHING`.

Once those two are done, this is good to merge from my POV.

DilumAluthge (approved):

PR #422, opened 2025-01-30
Prevent all "launch jobs" from running on pibots

DilumAluthge:
This is a follow-up to https://github.com/JuliaCI/julia-buildkite/pull/414


staticfloat:
I understand why you're doing this, but it would be nice to retain non-x86_64 workers for running these jobs.

What do you think about modifying the [environment variable configuration here](https://github.com/JuliaCI/sandboxed-buildkite-agent/blob/8e71aba279eb63d8be0027c0d225ded45ee3530d/linux-sandbox.jl/common.jl#L267-L276) to include a [mapping for `BUILDKITE_PLUGIN_JULIA_ARCH`](https://github.com/JuliaCI/julia-buildkite-plugin/blob/b4d0bfa510bc399417133adae013385af5c71190/hooks/pre-command#L107).  This would allow us to provide a good default value for the whole agent, which I think is probably a good idea anyway.

IanButterworth:
I guess this would fix why we get
Screenshot 2025-03-08 at 11 24 32 PM
https://buildkite.com/julialang/julia-master/builds/45605#019578ff-d9eb-4ff7-9779-d7be89d5c3fe/925-927

giordano:
@staticfloat can we go on with this PR? It'd of course be better to be able to use as many agents as possible, but the fact is that at the moment these jobs keep failing quite frequently because there are still some agents which can't work (e.g. the powerpc ones), so current situation is far from ideal anyway and quite frustrating for the developers.

staticfloat:
Yes, we can. It’s just a shame that this disables the aarch64 builders too, as they do a lot of these builds as well. Let’s just do this and fix it later when we have nice things like armv7l builds and ppc builds working.

giordano:
Can arch be made into a list without too much effort? At least that'd allow having both x86_64 and aarch64

staticfloat:
I think the way to do this is to add a tag to those two runner types.

staticfloat:
I think the way to do this is to add a tag to those two runner types.

IanButterworth:
> Let’s just do this and fix it later

I'm following this directive :)

Issue #429, opened 2025-03-09
No test report upload job for several test jobs


IanButterworth:
- test i686-linux-gnunet
- test x86_64-linux-gnuassertrr
- test x86_64-linux-gnuassertrr-net
- test x86_64-linux-gnummtk
- test x86_64-linux-musl

PR #430, opened 2025-03-09
Fix job id meta-data var name

IanButterworth:
Otherwise two different jobs were saving to `BUILDKITE_TEST_JOB_ID_test_i686-linux-gnu`.

The upload jobs still just use `TRIPLET` but I think that's ok as the `gnunet` one currently doesn't have reports uploaded.
i.e. #429


IanButterworth:
Incorporated into #433

PR #431, opened 2025-03-09
Standardize on Julia v1.7.3 for launching jobs

staticfloat:
This allows us to use the pibots, at least until we fix them and are able to build newer julias for them.

As discussed on slack with @DilumAluthge


DilumAluthge (approved):

IanButterworth:
I think this is now hitting a ppc issue?
Screenshot 2025-03-09 at 5 04 06 PM

giordano:
I guess the question is: what's the latest version of julia available for all platforms we have runners for?

PR #428, opened 2025-03-06
Assign test results back to their test job

IanButterworth:
Depends on
- [x] https://github.com/JuliaLang/julia/pull/57632
- [ ] https://github.com/buildkite-plugins/test-collector-buildkite-plugin/pull/84

## Motivation

Test results were previously associated with the upload jobs, so quite hidden.

Screenshot 2025-03-06 at 12 18 04 PM

Screenshot 2025-03-06 at 12 18 09 PM

PR #433, opened 2025-03-09
Upload results.json using raw api, rather than test-collector

IanButterworth:
Redo of #428
Incorporates #430 

It's too difficult to use `test-collector` to upload these files with a custom job-id, and #428 did it in an unsafe way.


pipelines/main/misc/upload_buildkite_results.yml
```diff
@@ -16,14 +16,38 @@ steps:
...
+          for file in results*.json; do
+            echo "Uploading $$file..."
+            # We can't use the test-collector plugin because it doesn't provide a way to set a custom `job_id`
+            curl \
+              -X POST \
+              --silent \
+              --show-error \
+              --max-time "${TIMEOUT?}" \
+              -H "Authorization: Token token=\"$$BUILDKITE_ANALYTICS_TOKEN\"" \
+              -F "data=\"@$$file\"" \
```
IanButterworth, line 38:
```suggestion
              -F "data=@\"$$file\"" \
```

utilities/test_julia.sh
```diff
@@ -160,8 +160,11 @@ else
 fi
 
 echo "--- Upload results.json report"
+# store the test job id so that the upload job can assign the results to the right job id
+buildkite-agent meta-data set BUILDKITE_TEST_JOB_ID_${BUILDKITE_STEP_KEY} "${BUILDKITE_JOB_ID}"
+echo "meta-data BUILDKITE_TEST_JOB_ID_${BUILDKITE_STEP_KEY} has been set to \"$(buildkite-agent meta-data get BUILDKITE_TEST_JOB_ID_${BUILDKITE_STEP_KEY})\""
 if compgen -G "${JULIA_INSTALL_DIR}/share/julia/test/results*.json"; then
-    (cd "${JULIA_INSTALL_DIR}/share/julia/test"; buildkite-agent artifact upload results*.json)
+    (cd "${JULIA_INSTALL_DIR}/share/julia/test"; buildkite-agent artifact upload "results*.json")
```
IanButterworth, line 167: Fixes this from https://github.com/JuliaLang/julia/pull/57686
```
2025-03-09 20:18:09 INFO   Found 1 files that match "results_ArgTools_1.json"
fatal: failed to upload artifacts: uploading artifacts: creating uploader: invalid upload destination: 'results_Artifacts_1.json'. Only s3://*, gs://*, rt://*, or https://*.blob.core.windows.net destinations are allowed. Did you forget to surround your artifact upload pattern in double quotes?
```

IanButterworth:
@staticfloat is this ok? One regression is the plugin did parallel uploads, isn't an issue on julia master because we only have one results file, but with https://github.com/JuliaLang/julia/pull/57686 we will have many.

I was thinking of merging this and testing it on that PR to see if I need to add the parallel uploads back?

vchuravy (approved):

vchuravy:
The only thing would be to open an issue with the test collector plugin and ask for support of a custom `job_id`

IanButterworth:
> The only thing would be to open an issue with the test collector plugin and ask for support of a custom job_id

Already open https://github.com/buildkite-plugins/test-collector-buildkite-plugin/pull/83
And a PR https://github.com/buildkite-plugins/test-collector-buildkite-plugin/pull/84 but the PR doesn't work because the env in the pre-exit process doesn't see new env vars from the previous steps (and sourcing the env var file proved both a bad idea)

staticfloat:
Belated LGTM!

PR #447, opened 2025-05-01
DO NOT MERGE

LebedevRI:
As suggested in https://github.com/JuliaLang/julia/pull/58284#issuecomment-2844346668 by @Zentrik, testing a `deps` change that was previously reverted.


LebedevRI:
@Zentrik @giordano could you please help me interpret the CI output here?
Given that the baseline CI isn't green, it is not obvious to me what the result is.

Zentrik:
The specific job is
https://buildkite.com/julialang/julia-buildkite-scheduled/builds/1274#01968d3b-e44b-4905-97b2-f54d5732b267.
Not immediately clear why it failed.

PR #450, opened 2025-06-03
Fix misc shellcheck complaints and run shellcheck on CI

fredrikekre:


fredrikekre:
This works now. It caught some problematic issues so I think it is worth adding shellcheck to CI like here.

fredrikekre:
Can I just merge or does it have to go through the mergebot in order to not mess with the signatures?

PR #453, opened 2025-06-30
Add Windows version of `--trim` tests

topolarity:
This will add Windows CI coverage for `--trim` once https://github.com/JuliaLang/julia/pull/57481/ lands.


IanButterworth:
> The Linux trimming job is failing here.

And the windows one. I'll convert this to draft to indicate it appears to need work. (rather than keep trying to get CI green)

PR #458, opened 2025-07-29
Disable musl

IanButterworth:
Build is very likely to hit https://github.com/JuliaLang/julia/issues/52707
Then we're now hitting a libbz2 issue that needs a fix on the musl upstream https://github.com/JuliaLang/julia/issues/40556

So it seems like a waste of CI resources?


DilumAluthge (approved):

ii8:
Is there any chance musl binaries will come back at some point? Otherwise I'll be stuck on 1.10 I guess.

I realise julia makes some funky use of dlopen that doesn't always work as expected on musl systems (or anything that isn't glibc I expect,) but I've simply stuck to using pure julia libraries only and was able to use julia fine this way.

matbrgz:
Adding a downstream use case here, not to argue with the decision — the rationale in this PR and in JuliaLang/julia#52707 / #40556 is clear enough, and we are not asking for the builder to come back.

We maintain [microHelium](https://github.com/UnitedOpen-Source/microHelium), a programming-contest judge (same family as DOMjudge/BOCA). It compiles and runs contestant submissions inside an Alpine Linux (musl) container, on aarch64, because the sandboxing story is much simpler for us on a small static-ish base image. We wanted to add Julia as a contest language, and could not.

What made it cost us time was not the outcome but the discovery path. From a user's side the only visible signal is `versions.json`, where `x86_64-linux-musl` simply stops appearing after 1.12.0-rc1:

```
versions with x86_64-linux-musl: 80
newest one: 1.12.0-rc1
```

and the tarballs 404 (`bin/musl/x64/1.12/julia-1.12.0-musl-x86_64.tar.gz`), while the 1.11.6 one still returns 200. The [support-tier page](https://julialang.org/downloads/support) says Tier 3, "binaries may be available in some cases", which reads as intermittent rather than discontinued — and it genuinely was intermittent for a while, since musl is also missing from 1.11.0, 1.11.1, 1.11.3, 1.11.4 and 1.11.7–1.11.9, so 1.11.6 is the last stable release with a musl binary. Nothing in the 1.12 `NEWS.md`/`HISTORY.md` mentions it. We only landed on this PR after working backwards through the buildkite config.

So the one cheap thing that would have helped is a line on the download/support page saying the builds were discontinued and pointing here. I opened JuliaLang/www.julialang.org#2598 for exactly that and offered to send the PR; flagging it here mostly so @ii8's question above has an answer in the thread: the builder was deliberately disabled in this PR, and the last musl binary published is 1.12.0-rc1.

For anyone else arriving here from a container: the practical options are a glibc base image, or building from source on musl, which upstream notes is unlikely to pass tests. We are going the first route.

PR #465, opened 2025-08-02
Improve timeout scripts to capture more information

Keno:
- Switch rr ones back to SIGTERM, to actually terminate recording gracefully
- On Linux, when SIGQUIT'ing walk the process tree to try to get coredumps from each stuck process, rather than just the toplevel (which is often just waiting for a worker).

Written by Claude.

### DIFF
--- a/utilities/timeout.jl
+++ b/utilities/timeout.jl
@@ -17,6 +17,14 @@
 
 include(joinpath(@__DIR__, "proc_utils.jl"))
 
+# Check if --term option is passed
+use_sigterm = false
+args_to_pass = ARGS
+if length(ARGS) > 0 && ARGS[1] == "--term"
+    use_sigterm = true
+    args_to_pass = ARGS[2:end]
+end
+
 # Parse a time period such as "2h30m"
 function parse_time_period(period::AbstractString)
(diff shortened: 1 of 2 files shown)


utilities/timeout.jl
```diff
@@ -60,9 +68,86 @@ timer_task = @async begin
...
-        kill(proc, Base.SIGQUIT)
-        println(stderr, "\n\nSent SIGQUIT to PID $(proc_pid).")
+        if use_sigterm
+            println(stderr, "\n\nProcess failed to exit within $(term_timeout)s, requesting termination (SIGTERM) of PID $(proc_pid).")
+            kill(proc, Base.SIGTERM)
+            println(stderr, "\n\nSent SIGTERM to PID $(proc_pid).")
+        else
+            println(stderr, "\n\nProcess failed to exit within $(term_timeout)s, requesting termination and coredump (SIGQUIT) of PID $(proc_pid).")
+            
+            # On Linux, recursively find and signal all descendant processes
```
staticfloat, line 78: I think a better solution would be to change the launch to put the processes in their own process group, then you can SIGQUIT the entire group at once.

PR #469, opened 2025-08-02
Try running tests in their own session id

Keno:
Alternative to https://github.com/JuliaCI/julia-buildkite/pull/465. I think it might be even better to just do `setpgid(0, 0)`, but libuv doesn't support that natively. Let's see if `setsid` works.


staticfloat (approved):
LGTM

Keno:
Well, the macos test hung, but ironically it showed that this is in fact working, so I'll merge this.

PR #471, opened 2025-08-05
coverage job: build julia with sysimage coverage

IanButterworth:
- Significantly speed up coverage tests, which have been timing out making it look like we have bad coverage.
- Enable coverage on macOS and Windows too
- There's a fair amount of duplication in the yml with this, but I think we should proceed and someone who's more familiar with our buildkite setup could DRY it all?


pipelines/scheduled/coverage/coverage.yml
```diff
@@ -5,7 +5,7 @@ steps:
         key:   "coverage"
         # Comment out the following `if:` statement for testing.
         # Please remember to uncomment it before merging the PR.
-        if: pipeline.slug == "julia-master-scheduled"
+        if: (pipeline.slug == "julia-buildkite") || (pipeline.slug == "julia-master-scheduled")
```
Keno, line 8: `julia-buildkite-scheduled`?

IanButterworth: bingo

IanButterworth:
Sometimes these coverage jobs fail on macOS because `Darwin honeycrisp-HL2F7YQ3XH` has python 3.12+ where they removed the distutils stdlib.

A fix is here https://github.com/JuliaCI/julia-buildkite-plugin/pull/59

IanButterworth:
Quoting @vtjnash on slack:

> code-coverage is a runtime flag and should never be used to build an image (it has no effect on it)


So while this _works_ the drop from 85 to 60-70% is likely mostly because the native code in the sysimage doesn't have tracking enabled. 

Plan for a proper fix discussed here: https://github.com/JuliaLang/julia/issues/57668#issuecomment-3156748844

IanButterworth:
Pivoting this to trying it out with https://github.com/JuliaLang/julia/pull/59234

IanButterworth:
I'm proposing merging this, then continuing on to fix the issues outlined in https://github.com/JuliaCI/julia-buildkite/issues/473

IanButterworth:
2 of these failed because of internet outage of some sort. The windows one timed out at 12h. 

All have passed before this with shorter run times so I think it's worth keeping them all to experiment.

PR #474, opened 2025-08-14
Upload to R2

fredrikekre:


DilumAluthge:
Just to clarify, for the time being, we're going to upload to both S3 and R2, right?

fredrikekre:
I need to update the aws_uploader rootfs, it has a very old aws cli version.

PR #475, opened 2025-08-15
Fix coverage job - Part 2

### LINKED ISSUE #473: Need to move from Coverage.jl to CoverageTools.jl
As @vtjnash pointed out, Coverage.jl is incorrect today and CoverageTools.jl is the correct option.

IanButterworth:
Fixes #473 

- Uses the new Coverage.jl API https://github.com/JuliaCI/Coverage.jl/pull/349
- Handles parallel jobs in the way that coveralls wants us to
- Fixes collecting coverage for the Compiler julia dir


pipelines/scheduled/coverage/coverage.yml
```diff
@@ -179,3 +179,33 @@ steps:
           JULIA_SHELL: "/bin/bash"
           # Receive cryptic token from parent job
           BUILDKITE_PLUGIN_CRYPTIC_BASE64_SIGNED_JOB_ID_SECRET: ${BUILDKITE_PLUGIN_CRYPTIC_BASE64_SIGNED_JOB_ID_SECRET?}
+
+      - label: ":white_check_mark: Finalize parallel coverage"
```
IanButterworth, line 183:
```suggestion
      - label: "Finalize parallel coverage"
```
claude loves sneaking emoji in

IanButterworth:
Todo:
- skip all ext stdlibs given we throw coverage away for them at the end
- figure out what to do on macOS given the new Coverage approach isn't viable there for the coveralls installer
- try to characterize why windows is so much slower, and fix

IanButterworth:
We could disable Profile and cmdlineargs tests in the Windows coverage job. That might be enough

PR #487, opened 2025-09-26
Revert "Use `-Werror` (with exceptions) on Windows builds"

ararslan:
Reverts JuliaCI/julia-buildkite#484 per https://github.com/JuliaCI/julia-buildkite/pull/484#issuecomment-3340764413.

We can't build 1.12.0-rc3 binaries for Windows due to `-Werror`.

PR #496, opened 2025-10-17
[re-land] Use `-Werror` (with some exceptions) on Windows builds

topolarity:
Re-land https://github.com/JuliaCI/julia-buildkite/pull/484/ (as requested in https://github.com/JuliaCI/julia-buildkite/pull/487#issuecomment-3394918901)

This will need the treehashes to be signed again.


DilumAluthge:
CI seems unhappy for some reason - I'm trying to figure out why.

PR #518, opened 2026-02-21
test_revise: use jlutilities manifest from testee

Keno:
Use the pinned Revise manifest from deps/jlutilities/revise/ when available, falling back to Pkg.add(rev="master") for older Julia versions that don't have it. The idea here is to be able to try Revise (/deps) PRs together with the base commits that necessitate them.

### DIFF
--- a/pipelines/main/misc/test_revise.yml
+++ b/pipelines/main/misc/test_revise.yml
@@ -30,7 +30,11 @@ steps:
 
           echo "--- Install and test Revise"
           unset JULIA_DEPOT_PATH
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Hm. It should be.

Keno:
```Activating project at `/cache/build/builder-amdci4-0/julialang/julia-buildkite/deps/jlutilities/revise` ```

Ok, this part is working, but looks like that manifest may be outdated - let me submit a separate base PR to update it and then we'll use it going forward.

PR #520, opened 2026-02-25
Upload `dmesg` log for failed Unix builds (e.g. to look for evidence of the OOM killer)

Keno:
The idea is that if e.g. the OOM killer is active, we can see evidence of that in the dmesg logs. Written by Claude.

### DIFF
--- a/utilities/test_julia.sh
+++ b/utilities/test_julia.sh
@@ -210,4 +210,12 @@ else
 fi
 echo "--- Done"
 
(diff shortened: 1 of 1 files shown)


DilumAluthge (commented):
One question. Apart from that, LGTM.

Also, the `shellcheck` CI job is green, which is good.

DilumAluthge (approved):

PR #517, opened 2026-02-15
Changing MMTk CI to use StickyImmix as default

udesou:

PR #541, opened 2026-05-18
Change MMTk CI to use StickyImmix

qinsoon:
This PR changes the GC plan to run for MMTk from Immix to StickyImmix (generational).


qinsoon:
This PR is duplicated from https://github.com/JuliaCI/julia-buildkite/pull/517/changes, as I don't have permissions to push to that PR.

qinsoon:
The MMTk build is failing. I am waiting for https://github.com/JuliaLang/julia/pull/61830 to be merged.

qinsoon:
https://github.com/JuliaLang/julia/pull/61830 has now been merged, so I expect MMTk with Sticky Immix should work correctly.

For the CI checks in this PR, there was one failing test under MMTk. However, the same test also failed in the run using Julia’s stock GC, so it does not appear to be specific to MMTk.

How would you like to proceed? @gbaraldi

topolarity:
This now also makes MMTk (StickyImmix) a required build job - no longer allow-fail.

@Keno let me know if that seems premature to you. I figured this might be a good point to make the cut.

Keno:
Ok from my perspective, I think.

topolarity:
All green for MMTk!

I have opted to keep this `allow-fail` for a couple weeks so that we can watch its health.
Only known unresolved bug is an OOM that seems to be aggravated by the recent CI changes at least.

PR #542, opened 2026-06-11
Launch unsigned jobs only

gbaraldi:


maleadt:
x-ref https://github.com/JuliaCI/sandboxed-buildkite-agent/pull/120, which disables cryptic.

PR #543, opened 2026-06-11
Split into build and test queue.

maleadt:


maleadt:
CI is currently entirely broken anyway, so let's merge this to test it out. If it turns out problematic, we can always revert. No easy way to try this out.

PR #547, opened 2026-06-13
Bump Julia plugin for cygwin fix.

maleadt:


maleadt:
I'm not particularly happy with this workaround, but let's merge it to unblock CI on Julia.

PR #548, opened 2026-06-13
test_julia: skip JuliaLowering_stdlibs under rr

maleadt:
The `gnuassrtrr` test job (x86_64-linux-gnuassert, run under rr) times out at its 6h Buildkite limit. The dominant cost is the JuliaLowering_stdlibs test, which precompiles every stdlib in two cache configurations using the pure-Julia JuliaLowering frontend. That workload is embarrassingly parallel, but rr records deterministically and serializes it onto effectively one core, so it took 2h45m in build 58055 (vs ~3min without rr) and was the long pole gating the entire parallel test phase, pushing the job past its timeout.

### DIFF
--- a/utilities/test_julia.sh
+++ b/utilities/test_julia.sh
@@ -83,8 +83,8 @@ if [[ "${USE_RR-}" == "rr" ]] || [[ "${USE_RR-}" == "rr-net" ]]; then
     export NCORES_FOR_TESTS="parse(Int, ENV[\"JULIA_RRCAPTURE_NUM_CORES\"])"
     export JULIA_NUM_THREADS=1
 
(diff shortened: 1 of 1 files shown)


Keno:
I don't specifically mind the exclusion, but

> but rr records deterministically and serializes

We have the ability to separate executions into separate rr traces (which we use for tests). We should hook that up.

maleadt:
I'll open an issue to keep track and investigate, but I'd like to expedite merging this because it's consuming quite some test resources. Not an issue in the weekend, but it's going to hog the Linux testers once real load hits.

(and once things have settled/recovered I'll ease on the merge button here)

PR #561, opened 2026-07-09
disable JuliaC pipeline for now

KristofferC:
Ref https://github.com/JuliaLang/julia/pull/62222#issuecomment-4914860627

### DIFF
--- a/utilities/render_launch_pipeline.py
+++ b/utilities/render_launch_pipeline.py
@@ -484,7 +484,9 @@ def main():
         )
 
     # JuliaC: itself a launcher with its own group + notify -- include verbatim.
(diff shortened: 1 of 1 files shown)


DilumAluthge (approved):
SGTM, but maybe open an issue (on this repo) and assign Cody and Gabriel? So we remember to re-enable it?

topolarity (changes requested):
I think I'd like to try to change my mind here (sorry)

There are some major serialization-facing changes incoming (e.g. https://github.com/JuliaLang/julia/pull/62245 and also several follow-ons) that I don't think can wisely land without `--trim` coverage in CI. Give me 24 hours to try to sort out / merge https://github.com/JuliaLang/julia/pull/61983/ so that we can avoid having to do this.

Apologies for the thrash!

PR #564, opened 2026-07-12
Port OIDC/KMS trust architecture to release-julia-1.13

Keno:
Bring the post-#544 CI architecture to the release-1.13 branch so that release-branch PRs run under the shared `julia-pr` pipeline and release nightlies run under `julia-ci` -> `julia-publish`, instead of the old per-release pipeline slugs (`julia-release-1-dot-13`, `julia-master-scheduled`, `julia-buildkite`). Per-release differences are now expressed as branch *content* selected by julia's in-repo `.buildkite-external-version` pointer, not by runtime `pipeline.slug` checks -- so no separate Buildkite pipelines are needed (the shared `julia-ci`/`julia-publish` already branch-limit to `master release-*`).

What changed:
  * Retire the cryptic secret-management + signed/unsigned pipeline split (cryptic_repo_keys/, secrets/, *.signature, launch_signed_jobs.yml, launch_unsigned_*.yml, launch_upload_jobs.yml, platforms/upload_*.yml).
  * Adopt main's OIDC + AWS KMS trust infrastructure, the single-upload render_launch_pipeline.py launcher, and the trusted julia-publish / julia-publish-test pipelines. The trailing `wait` + `trigger: julia-publish` is gated on `pipeline.slug == "julia-ci"`.

Preserved as release-1.13-specific:
  * The build/test matrix: every pipelines/**/platforms/*.arches file is kept verbatim from release-julia-1.13 (rootfs v6.00 / docker v7.10, mmtk on the `package_linux` image, Windows CFLAGS), plus launch_powerpc.jl.
  * The job set: render_launch_pipeline.py is trimmed to omit the JuliaSyntax and JuliaC launch groups (new on main; not part of release-1.13 CI). The juliasyntax.* / juliac/ YAMLs remain in-tree so either can be re-enabled by re-adding its `blocks.append(...)`.

NOT yet validated on Buildkite -- needs a CI run under the shared pipelines.

PR #544, opened 2026-06-12
Replace cryptic with Buildkite OIDC + AWS KMS trust infrastructure

Keno:
A little bit of Claude-slop, but I think generally in the right direction. The core idea is to store all key material in AWS KMS and use OIDC to mint step-appropriate IAMs that provide the necessary access for a particular step.

---

Fixes https://github.com/JuliaLang/julia/issues/62269

PR #565, opened 2026-07-12
release-julia-1.13-new: build matrix + job-set specialization

Keno:
This is the sole release-1.13-specific commit layered on top of `master`. The release-julia-1.13 branch is now `master` + this commit, so the shared OIDC/KMS CI architecture (#544) stays byte-identical to master and future master -> release-* cherry-picks apply cleanly, per review feedback on #564.

Everything release-1.13-specific lives here:

  * Build/test matrix (pipelines/**/platforms/*.arches): pin the rootfs images to v6.00 / docker v7.10, put mmtk on the `package_linux` image, and drop master's Windows -Wno-error CFLAGS -- i.e. keep release-1.13's existing build environment rather than master's (1.14) one.

  * Job set (utilities/render_launch_pipeline.py): omit the JuliaSyntax and JuliaC launch groups, which are new on master and not part of release-1.13 CI. The juliasyntax.* / juliac/ YAMLs stay in-tree (from master), so either group re-enables by re-adding its `blocks.append(...)`.

NOT yet validated on Buildkite.

### DIFF
--- a/utilities/render_launch_pipeline.py
+++ b/utilities/render_launch_pipeline.py
@@ -472,19 +472,15 @@ def main():
         allow_fail_group_text(),
     ]
 
-    # JuliaSyntax: gated on ./JuliaSyntax/Project.toml existing in the julia
-    # checkout (same conditional the old Check step used). Included verbatim
-    # as its own group (it is itself a launcher with its own notify).
-    juliasyntax_project = os.path.join(os.getcwd(), "JuliaSyntax", "Project.toml")
-    if os.path.exists(juliasyntax_project):
-        blocks.append(verbatim_group_text(os.path.join(MISC, "juliasyntax.launch.yml")))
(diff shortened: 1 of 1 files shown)


Keno:
@DilumAluthge See if you like this better. That said, we may need to bump the rootfs images anyway if we want the `aws` upload to work on julia-pr, but let's start here.

Keno:
Ok, so my plan is merge this right away (it's a new branch), try it out on the PR, if it works, swap the branch names.

DilumAluthge:
Sounds good. Then future commits to master can easily be cherry-picked ("forward ported", I suppose) to the release-1.13 branch.

I wouldn't delete the old release-1.13 branch, just rename it to something like `release-Julia-1.13-old`, so that we keep the git history around.

PR #566, opened 2026-07-14
Add JuliaLowering compatibility smoke jobs

aviatesk:
JuliaLang/julia#62367 restores enough compatibility for JETLS to load and precompile JuliaLowering on Julia 1.12, although currently there is no system to detect future regressions, as discussed in the PR.

So this commit adds Linux x86_64 matrix jobs for Julia 1.12 and 1.13 that load JuliaLowering with `JULIA_LOWERING_PRECOMPILE=true`, without persisting compiled caches, ensuring each job exercises precompilation. This also registers the new JuliaLowering group in the launch pipeline renderer.

Note that the full test suite executes lowered code and requires newer runtime support, so it cannot be used for this minimum compatibility coverage.


topolarity (approved):
LGTM, let's merge if CI comes back green

PR #570, opened 2026-07-20
macOS codesign: sign each inode once so hardlinked Mach-Os notarize

Keno:
Julia's .app bundles julia's launcher as a hardlink: contrib/mac/app's rule does `ln -f .../julia .../julia-terminal`, so both paths share one inode. codesign.sh signs the tree with a parallel pool (find -type f -perm -0111, up to MACOS_CODESIGN_JOBS workers), and rcodesign signs a Mach-O IN PLACE. Two workers therefore mutate the same inode concurrently and tear the signature, and Apple's notary rejects both paths:

  "The signature of the binary is invalid."
    .../Julia-1.14.app/Contents/Resources/julia/bin/julia
    .../Julia-1.14.app/Contents/Resources/julia/bin/julia-terminal

(Every julia-publish nightly since the launcher hardlink landed fails this way.) De-duplicate by inode in the signing loop: sign each inode once; the remaining hardlinks to it already carry that one valid signature (rcodesign signs in place, so the shared inode is updated for all of them). The bundle seal then covers the identical bytes at every path, and notarization passes.

### DIFF
--- a/utilities/macos/codesign.sh
+++ b/utilities/macos/codesign.sh
@@ -91,6 +91,16 @@ is_macho() {
     esac
 }
 
+# Inode number of $1, portable across GNU coreutils `stat -c`, BSD/macOS
+# `stat -f`, and a `ls -di` fallback. Used to sign each physical inode only
+# once: the bundle contains hardlinked Mach-Os (see the signing loop), and a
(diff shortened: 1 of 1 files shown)


Keno:
We're not running this code on this repo's CI - so let me merge it and trigger an upstream build to see if it works.

PR #571, opened 2026-07-21
macOS notarize: break julia/julia-terminal hardlink before signing

Keno:
Supersedes the inode-dedup in 205ae8f, which was the wrong layer: a julia-publish run (build #59) with that change still failed notarization with the identical error on both binaries --

  "The signature of the binary is invalid."
    Julia-1.14.app/Contents/Resources/julia/bin/julia
    Julia-1.14.app/Contents/Resources/julia/bin/julia-terminal

even though the dedup demonstrably ran ("Skipping hardlink to already-signed inode ..."). So the failure is not a concurrent-signing race.

Root cause is the shared inode itself. contrib/mac/app builds the launcher as a hardlink -- bin/julia-terminal and bin/julia are one inode with two names -- and the bundle's main executable is julia-terminal (via the Contents/MacOS/julia-terminal symlink). The bundle seal signs that inode with the MAIN-EXECUTABLE signature, which therefore also lands on bin/julia. Apple's notary validates bin/julia as an ordinary nested binary, finds a main-executable signature, and rejects both paths. (160 non-hardlinked Mach-Os in the same bundle notarize fine; only the hardlink pair fails.)

Fix: in upload_julia.sh, split the hardlink into two independent regular files before signing, so each gets its own correct signature. A full copy keeps the name-based loader dispatch (cli/loader_exe.c) and the bin/ rpath, exactly like the hardlink; bin/julia is the small loader so the extra copy is negligible. codesign.sh is reverted to its pre-205ae8f form.

### DIFF
--- a/utilities/macos/codesign.sh
+++ b/utilities/macos/codesign.sh
@@ -91,16 +91,6 @@ is_macho() {
     esac
 }
 
-# Inode number of $1, portable across GNU coreutils `stat -c`, BSD/macOS
-# `stat -f`, and a `ls -di` fallback. Used to sign each physical inode only
-# once: the bundle contains hardlinked Mach-Os (see the signing loop), and a
-# single filesystem holds the whole tree, so the inode alone identifies a file.
-inode_of() {
-    stat -c '%i' -- "${1}" 2>/dev/null \
-        || stat -f '%i' -- "${1}" 2>/dev/null \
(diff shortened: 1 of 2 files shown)


Keno:
This is one of those things that's easiest to debug in production because we need to run against the real apple notarization service, so ... sorry for the churn.

PR #515, opened 2026-02-01
Upload the `x86_64-linux-gnu` asserts build to S3

DilumAluthge:

PR #576, opened 2026-07-25
1.13: Upload the `x86_64-linux-gnu` asserts build to S3 (#515)

wsmoses:
(cherry picked from commit 0d1a26f704569fe8a4f3c3d399e9e8d1c1e8466a)

### DIFF
--- a/pipelines/main/platforms/upload_linux.arches
+++ b/pipelines/main/platforms/upload_linux.arches
@@ -2,6 +2,7 @@
 aarch64-linux-gnu          .
 i686-linux-gnu             .
 x86_64-linux-gnu           .
+x86_64-linux-gnuassert     .
 
 # These special lines allow us to embed default values for the columns above.
 # Any column without a default mapping here will simply substitute a `.` to the empty string


DilumAluthge:
I think you want to target `release-julia-1.13-new`

wsmoses:
done!

DilumAluthge:
Just FYI, I renamed `julia-buildkite#release-julia-1.13-new` to ``julia-buildkite#release-1.13`, which is what you'll see now.

Keno:
The -old pipelines are not used anymore and the new pipeline doesn't have a separate upload step.

DilumAluthge:
Ah. So should I revert #515?

And then should I delete the `pipelines/main/platforms/upload_linux.arches` file off of `main`?

wsmoses:
in which case @Keno what's the right way we should make sure the assert builds from releases are uploaded?

Keno:
I don't know that you need to revert, but yeah, I think the upload jobs can be deleted.

Keno:
> in which case @Keno what's the right way we should make sure the assert builds from releases are uploaded?

They are uploaded, e.g.  `s3://julialang-ephemeral-ci/bin/5621f8ed4400f063dde3ff7ef19bf1e19a6d5baa/julia-5621f8ed44-linuxassert-x86_64.tar.gz`. Do you need them code-signed etc?

wsmoses:
But those are only ephemeral right? We're thinking about releases, so i can get say 1.12.7 assert builds downloaded for use in ci

Keno:
Ok, the upload_ builds don't help with that either though - they just used to move buildkite -> ephemeral S3. If you want proper releases, that needs to be added to the publish list.

Keno:
(which I guess does still use the upload arch list, so maybe this is working after all and we just need to rename the upload_ files now).

Keno:
That said, you don't need to do it on each branch. For security, since it has signing credentials, there's only one global publish- pipeline that's pinned in the webui.

DilumAluthge:
> Do you need them code-signed etc

I believe we only need Linux assert builds uploaded, so code-signing not really relevant (GPG-signing would be nice, but I don't think is required for this use case) - @wsmoses correct me if I'm wrong?

Keno:
Well, I think if they're in the release buckets it would be awkward if they didn't go through the same process as the other binaries.

wsmoses:
Yeah we just want the prebuilt binaries for each release to be available for us to download in ci (or ideally later on use in the setup julia action) vs right now have to build from source every commit (as we've run out of disk cache across the repo to meaningfully persist things)

DilumAluthge:
> Well, I think if they're in the release buckets it would be awkward if they didn't go through the same process as the other binaries.

I'm fine putting them through the same process, if that's easiest.

DilumAluthge:
> that needs to be added to the publish list

@Keno Can you remind me - where is the publish list? Is it in the web UI, or in this repo?

Keno:
https://github.com/JuliaCI/julia-buildkite/blob/main/utilities/publish.sh#L22-L27 - although I'm not sure what our process is for building release-release binaries as opposed to just binaries on the release branch

wsmoses:
so then this PR is exactly what's needed, then?

wsmoses:
except seemingly not needing a backport if all julia releases just use the main version?

Keno:
Well, the pin in the webui needs to definitely be updated. Other than that, I don't think we've cut an actual release with the new pipeline yet, so I'm not entirely sure.

DilumAluthge:
I think there's probably enough stuff waiting on the `julia#release-1.10` branch. I can try to work on a new 1.10.x patch release.

DilumAluthge:
And then we can use that patch release to make sure CI works as expected on an actual "release release"

PR #577, opened 2026-07-25
Add julia-build-request: build a single commit on demand

Keno:
PkgEval needs a binary for whatever commit it is asked to evaluate. CI has usually already built it, and the staged artifact is fetched directly -- but not always: a closed PR, an expired ephemeral artifact, or a commit CI never built leave it with no option but to compile Julia itself, once per worker, which for a distributed run is hours of duplicated work.

Add an untrusted pipeline that builds exactly one commit and one variant, created only through the REST API (no push or PR triggers) by a machine user whose team access covers this pipeline alone.

It may be asked to build an arbitrary commit, which is safe for the same reasons julia-pr is: the commit is attested by Buildkite, the stage role can only write below its own attested commit, and -- as with the self-test CI -- it stages to its own bucket, so nothing juliaup or julia-publish reads can originate from a requested build.

Everything else follows the existing machinery: adding the slug to build_pipelines and s3_staging_buckets creates the bucket, its policies and the write-once stage role; build_envs.sh and aws_oidc.sh gain one branch each; the build step is the ordinary build_linux.yml template rendered for the single requested triplet.

Requires the pipeline UUID in buildkite_ids.auto.tfvars before apply.

### DIFF
--- a/ops/terraform/buildkite_ids.auto.tfvars
+++ b/ops/terraform/buildkite_ids.auto.tfvars
@@ -10,17 +10,24 @@
 buildkite_organization_id = "d409823c-5fa7-41c8-9033-7269c5fde4f3"
 
 buildkite_pipeline_ids = {
-  "julia-pr"           = "019ebd61-5b1f-428e-b08c-1b5a2111e001"
-  "julia-ci"           = "019ebd63-635e-4155-b60a-9d2815900786"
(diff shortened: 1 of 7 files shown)


Keno:
Looks like it's working: https://buildkite.com/julialang/julia-build-request/builds/5/canvas

PR #586, opened 2026-08-04
Fix release tag double build

Keno:
Turns out we build once for the branch and once for the tag. This makes those builds not collide, but possibly we should instead not build on tags at all and do something smarter.

### DIFF
--- a/utilities/build_envs.sh
+++ b/utilities/build_envs.sh
@@ -121,9 +121,58 @@ MAJMIN="$(cut -d. -f1-2 <<<"${JULIA_VERSION}")"
 MAJMINPAT="$(cut -d- -f1 <<<"${JULIA_VERSION}")"
 export JULIA_VERSION MAJMIN MAJMINPAT
 
-# If we're on a tag, then our "tar version" will be the julia version.
-# Otherwise, it's the short commit.
-if git describe --tags --exact-match >/dev/null 2>/dev/null; then
+# Is this job part of the RELEASE flow for a v* tag? True for the tag
+# build itself -- BUILDKITE_TAG is set uniformly on every job of a tag
+# build, and Buildkite reports the tag name as the branch, which also
+# covers a manual re-run created with branch=v<version> -- and for the
+# julia-publish run it triggers (the trigger passes the branch through,
+# pinned to the source build's by the cluster trigger rule). The
+# release-*/master build of the very same commit deliberately does NOT
+# count: the tag build is an independent build, and only ITS artifacts
+# are promoted to the versioned release locations (see STAGING_TARGET
+# below for why the two must not be interchangeable). Restricted to the
+# pipelines that feed releases so that e.g. a julia-pr branch that
+# happens to be named v2-something cannot enter the release flow.
+RELEASE_TAG_FLOW="false"
+case "${BUILDKITE_PIPELINE_SLUG:-}" in
+    julia-ci|julia-publish*)
+        if [[ "${BUILDKITE_TAG:-}" == v[0-9]* || "${BUILDKITE_BRANCH:-}" == v[0-9]* ]]; then
+            RELEASE_TAG_FLOW="true"
+        fi
+        ;;
+esac
+export RELEASE_TAG_FLOW
+
+# The release flow names everything after JULIA_VERSION (the VERSION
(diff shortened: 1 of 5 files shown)


utilities/upload_to_s3.sh
```diff
@@ -32,6 +34,16 @@ upload_to_s3() {
...
     fi
 
+    # Stamp every object with the commit that produced it. BUILDKITE_COMMIT
+    # is the same Buildkite-attested value that gates the staging paths and
+    # that verify_trusted_commit.sh checks before a publish, so the stamp is
+    # trustworthy provenance -- and it is what lets a second build of the
+    # SAME commit be treated as idempotent below.
+    local metadata_args=()
+    if [[ "${BUILDKITE_COMMIT:-}" =~ ^[0-9a-f]{40}$ ]]; then
+        metadata_args=( --metadata "build-commit=${BUILDKITE_COMMIT}" )
```
DilumAluthge, line 44: Just to clarify, is this Buildkite metadata or S3 metadata?

utilities/build_envs.sh
```diff
@@ -121,9 +121,58 @@ MAJMIN="$(cut -d. -f1-2 <<<"${JULIA_VERSION}")"
...
+# julia-publish run it triggers (the trigger passes the branch through,
+# pinned to the source build's by the cluster trigger rule). The
+# release-*/master build of the very same commit deliberately does NOT
+# count: the tag build is an independent build, and only ITS artifacts
+# are promoted to the versioned release locations (see STAGING_TARGET
+# below for why the two must not be interchangeable). Restricted to the
+# pipelines that feed releases so that e.g. a julia-pr branch that
+# happens to be named v2-something cannot enter the release flow.
+RELEASE_TAG_FLOW="false"
+case "${BUILDKITE_PIPELINE_SLUG:-}" in
```
DilumAluthge, line 137: I might be misunderstanding, but can't an attacker edit this file, and thus this logic?

DilumAluthge:
To clarify, I have one question about disambiguating the word "metadata", but apart from that, my only concerns about security - so if this is already secure (from our "separate pipelines" setup), then I think this is good to merge. The code itself seems fine.

(Although it would be nice to add a comment stating whether "metadata" is Buildkite metadata or S3 metadata. I'm looking at the code again, and I _think_ it's S3 metadata, but it would be nice to explicitly state in a comment, to make the code easier to read).

utilities/build_envs.sh
Keno: julia-ci runs trusted builds from master

utilities/upload_to_s3.sh
Keno: S3 metadata

utilities/build_envs.sh
DilumAluthge: Ah, so this code doesn't run in the `julia-pr` pipeline?

If so, then this PR LGTM to merge.

Keno: It does, but the julia-pr pipeline is OIDC restricted to only be able to upload it's particular release artifact.

DilumAluthge: I see, so this code will be run in `julia-pr`, but S3 will not permit `julia-pr` to upload to release locations, so even if an attacker changes this code, they don't suddenly gain the ability to upload to a release location. Is that correct?

Keno: Yes

DilumAluthge: Sounds good - ship it.

DilumAluthge (approved):
LGTM

DilumAluthge:
@Keno I don't have permission to rebase this PR. Can you enable the "allow maintainers to push to this PR" option in the PR?

giordano:
I don't think you can edit PRs coming from orgs.

DilumAluthge:
😭

DilumAluthge:
@Keno Maybe you could give me write permissions on the fork?

In the meantime, for time's sake, I'll cherry-pick this into a separate PR.

DilumAluthge:
Oh. Maybe I can merge directly by "bypassing" the branch protection rule. CI is all green, but the PR tip is behind master, so hopefully this doesn't break anything.

PR #592, opened 2026-08-11
ops: add a julia-promote pipeline for the release bucket dance

KristofferC:
Promoting a release from the nightlies bucket to the release bucket (julialang2, served at julialang-s3.julialang.org) has been a manual, undocumented release-manager step with ad-hoc AWS credentials. This adds a trusted **julia-promote** pipeline that does it as a single step: the release manager creates a New Build with the tag name as branch (after the tag's julia-publish run has passed), and the step:

- re-maps the published artifacts into the release bucket's historical layout (`linux/x86_64`→`linux/x64`, `i686`→`x86`, `freebsd/x86_64`→`freebsd/x64`; the mac/winnt short names map 1:1) — verified against what v1.13.0-rc1 looks like on julialang2
- repoints the `julia-<majmin>-latest` pointers
- generates and uploads `bin/checksums/julia-<version>.{sha256,md5}`
- purges the Fastly cache and verifies every object is reachable via the CDN

Trust posture mirrors julia-publish: manual/no-PR builds only, a dedicated `julia-oidc-promote` role gated on the pipeline UUID + `promote_*` step keys, write-once uploads (the majmin-latest pointers become the second overwrite exception in `upload_to_s3.sh`), and a provenance check that each promoted object's `build-commit` metadata (stamped by the publish role) matches the build's own commit — so a repointed tag cannot silently promote stale binaries. No KMS access; everything was already signed by publish.

One deliberate divergence: the pipeline loads its steps from julia-buildkite `main` instead of the julia checkout's `.buildkite-external-version` pin, because a promotion may target an older tag whose pin predates this pipeline, and the step runs no code from the julia tree (the checkout contributes only the attested commit + branch name).

**Deployment checklist** (admin with Buildkite + AWS access):
1. Create the `julia-promote` pipeline in the **Secure cluster**: webhook builds OFF, PR builds OFF, branch limiting `v*`; paste `pipelines/promote/0_webui.yml` as its steps.
2. Put its UUID in `buildkite_ids.auto.tfvars` (the placeholder zero UUID fails terraform validation, so this is fail-closed).
3. `terraform apply`.
4. First use: promote v1.13.0-rc2 by creating a build with branch `v1.13.0-rc2`, then dispatch the CI workflow on JuliaLang/VersionsJSONUtil.jl.

### DIFF
--- a/ops/README.md
+++ b/ops/README.md
@@ -43,6 +43,13 @@ only trust the publish pipeline's slug. There are three pipelines:
 - **`julia-ci`** — builds trusted refs only: master, release-*, tags, and the
   scheduled nightlies (untrusted to sign, but it is what triggers publish).
 - **`julia-publish`** — the trusted pipeline that signs + promotes.
+- **`julia-promote`** — the trusted pipeline that copies a published
+  release from the nightlies bucket into the release bucket
+  (julialang2, served at julialang-s3.julialang.org). No webhook or PR
+  builds; the release manager creates its builds manually (New Build
+  with branch = `v<version>`) after the julia-publish run for that tag
+  has finished. See `pipelines/promote/0_webui.yml` and
+  `utilities/promote_release.sh`.
 
 A fourth, untrusted pipeline, **`julia-buildkite-ci`**, is the
 julia-buildkite repository's own self-test CI (see `.buildkite/README.md`):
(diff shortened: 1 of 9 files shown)


Keno:
terraform deployed a rebased version of this

Keno (approved):

PR #601, opened 2026-08-13
publish: speed up the publish and docs-deploy jobs

KristofferC:
Wall-clock analysis of a `julia-publish` build (https://buildkite.com/julialang/julia-publish/builds/146) showed the `publish all` job dominated by serialized compression and notarization waits, plus the `upload docs` job holding a secure agent for a full-history clone of docs.julialang.org. This PR addresses all of it:

- **deploy_docs**: shallow-clone docs.julialang.org (`--depth 1`). The gh-pages history is multi-GB (one commit per deploy) and only the tip tree is needed to build the next commit; pushing from a shallow clone works because the new commit's parent is the remote tip.
- **Inno Setup**: `LZMANumBlockThreads=4` + `LZMAUseSeparateProcess=yes` — parallel lzma2 block compression, with the separate 64-bit compressor process avoiding the 32-bit compiler's memory ceiling.
- **upload_julia**: repack the release tarballs with `pigz` when available (gzip fallback; JuliaCI/rootfs-images#292 adds pigz to the `julia_publish` image), and `7z a -mmt=on` for the Windows zip.
- **Deferred dmg notarization**: `build_dmg.sh` only submits each dmg to Apple when running under `publish.sh`, which waits/staples/promotes them in a final phase — Apple's processing overlaps the remaining triplets instead of blocking each macOS one. Note: `rcodesign notary-wait` exits 0 even for a rejected submission (verified against the pinned 0.29.0 source), so the staple acts as the success gate (a rejected submission has no ticket to staple). Standalone invocation and `PUBLISH_SKIP_NOTARIZATION=1` keep the synchronous path.
- **verify_trusted_commit**: treeless fetches (`--filter=tree:0`) via a promisor remote — the reachability check only walks commit/tag objects. Verified locally against the real repo: the full check runs in ~6s. Servers without filter support ignore the filter, so this degrades gracefully.

Coverage note: the publish-test pipeline exercises everything except the deferred-notarization path (it sets `PUBLISH_SKIP_NOTARIZATION=1`), so that path first runs on a production publish.

Test build running at https://buildkite.com/julialang/julia-publish-test-nosecrets/builds/16)

### DIFF
--- a/pipelines/main/misc/deploy_docs.yml
+++ b/pipelines/main/misc/deploy_docs.yml
@@ -68,7 +68,10 @@ steps:
           echo "--- clone docs.julialang.org"
           mkdir -p /home/juliaci/.ssh
           ssh-keyscan github.com >> /home/juliaci/.ssh/known_hosts 2>/dev/null
-          git clone git@github.com:JuliaLang/docs.julialang.org -b gh-pages repo
+          # Shallow clone: only the tip tree is needed to build the next
+          # commit, and pushing from a shallow clone works because the new
+          # commit's parent is the remote tip.
+          git clone --depth 1 -b gh-pages git@github.com:JuliaLang/docs.julialang.org repo
           cd repo
 
           echo "--- deploy new docs"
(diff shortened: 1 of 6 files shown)


Keno (approved):
Sure. We can also run more of the things in parallel. That said, we do need to be a bit careful about AWS simultaneous upload limits - that's bitten us before in the past.

KristofferC:
We can reassess once this is in to see if further tweaks are needed.

PR #611, opened 2026-08-15
ci: add scheduled optimized Linux build

maleadt:
Build an x86-64 Linux nightly through Julia's PGO, ThinLTO, and BOLT stages, test it as allow-fail, and publish it under linuxopt. Use a generic stage-1 sysimage to keep profiling affordable while retaining the release CPU targets for stage 2.

Give the build path and scheduled publish explicit selectors. This leaves room for later optimized platforms without colliding with Julia's release/debug build mode or treating the regular-GPL optimized artifact as a no-GPL build.

Closes https://github.com/JuliaCI/julia-buildkite/pull/345
Closes https://github.com/JuliaCI/julia-buildkite/pull/492


maleadt:
The `gnuopt` build succeeded here, which is what we care about.

Zentrik (commented):
Lgtm, 🎉

